PR #1, opened 2024-03-26
Add contents

eemeli:
This adds the first set of functionality:
- Message and Resource dataclasses
- Parsers and serializers for:
  - Fluent, using `fluent.syntax`
  - .ini, using `iniparse` (which `translate-toolkit` uses internally)
  - .po, using `polib` (also tried `translate-toolkit` and `babel`, but this was better, and already used by Pontoon)
  - .properties, using `translate-toolkit`
- An `add_entries()` merge tool

I've broken up the whole of this (~3k lines in total...) into commits, which each add individual parts of the above.


eemeli:
Added DTD parsing & serialization. The former is based on [silme's](https://github.com/mozilla/silme/blob/master/lib/silme/format/dtd/parser.py), so it should not introduce any new bugs. The serialization is slightly improved, avoiding some additional edge cases.

eemeli:
Add `.inc` support. The format is still used by a few Seamonkey files.

moz/l10n/properties/serialize.py
```diff
@@ -0,0 +1,100 @@
...
+    encoding: Literal["iso-8859-1", "utf-8", "utf-16"] = "utf-8",
+    serialize_message: Callable[[V], str] | None = None,
+    serialize_metadata: Callable[[Metadata[M]], str | None] | None = None,
+    trim_comments: bool = False,
+) -> Iterator[str]:
+    """
+    Serialize a resource as the contents of a .properties file.
+
+    Section identifiers will be prepended to their constituent message identifiers.
+    Multi-part message identifiers will be joined with `.` between each part.
```
mathjazz, line 34: What is an example of a multi-part message identifier?

moz/l10n/properties/serialize.py
```diff
@@ -0,0 +1,100 @@
...
+
+    For non-string message values, a `serialize_message` callable must be provided.
+    If the resource includes any metadata, a `serialize_metadata` callable must be provided
+    to map each field into a comment value, or to discard it by returning an empty value.
+
+    Comment lines not starting with `#` will be prefixed with `# `.
+
+    Yields each entry, comment, and empty line separately.
+    Re-parsing a serialized .properties file is not guaranteed to result in the same Resource,
+    as the serialization may lose information about message sections and metadata.
```
mathjazz, line 44: What needs to be done to prevent this?

tests/test_properties.py
```diff
@@ -0,0 +1,269 @@
...
+                                comment="test whitespace trimming in key and value",
+                            ),
+                            Entry(
+                                ["3"],
+                                "\u1234\t\r\n\u00AB\\u0001\n",
+                                comment="test parsing of escaped values",
+                            ),
+                            Entry(
+                                ["4"],
+                                "this is multiline property",
```
mathjazz, line 159: Nice!

https://github.com/mozilla/pontoon/issues/2207

tests/data/test.properties
```diff
@@ -0,0 +1,13 @@
+# Any copyright is dedicated to the Public Domain.
+# http://creativecommons.org/publicdomain/zero/1.0/
+1=1
+ 2=2
+3 =3
+ 4 =4
+5=5
+6= 6
```
mathjazz, line 8: Remembered this one:
https://github.com/mozilla/pontoon/issues/2058

What do you think should be the next step for that issue?

mathjazz (commented):
Excellent work, really looking forward to it! Seems like we should close https://github.com/mozilla/pontoon/issues/2273 and file a similar issue. :)

Left some notes, nothing major. As mentioned, I'd like to test it with Pontoon a little.

README.md
```diff
@@ -0,0 +1,20 @@
...
+The core idea here is to establish [Message](./moz/l10n/message.py) and [Resource](./moz/l10n/resource.py)
+as format-independent representations of localizable and localized messages and resources,
+so that operations like linting and [transforms](./moz/l10n/transform/) can be applied to them.
+
+Parsers and serializers are provided for a number of formats,
+using common and well-established libraries to take care of the details.
+A unified API for these is provided,
+such that `FORMAT_parse(source)` will always accept `str` input,
+and `FORMAT_serialize(resource)` will always provide a `str` iterator.
+All the serializers accept a `trim_comments` argument,
```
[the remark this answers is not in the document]
eemeli:
```suggestion
All the serializers accept a `trim_comments` argument
which leaves out comments from the serialized result,
```

README.md
```diff
@@ -0,0 +1,20 @@
...
+primarily built for internal use at Mozilla.
+
+The core idea here is to establish [Message](./moz/l10n/message.py) and [Resource](./moz/l10n/resource.py)
+as format-independent representations of localizable and localized messages and resources,
+so that operations like linting and [transforms](./moz/l10n/transform/) can be applied to them.
+
+Parsers and serializers are provided for a number of formats,
+using common and well-established libraries to take care of the details.
+A unified API for these is provided,
+such that `FORMAT_parse(source)` will always accept `str` input,
```
eemeli, line 13:
```suggestion
such that `FORMAT_parse(text)` will always accept `str` input,
```

moz/l10n/fluent/__init__.py
```diff
@@ -0,0 +1,10 @@
+from .parse import fluent_parse, fluent_parse_message
+from .serialize import fluent_astify, fluent_astify_message, fluent_serialize
+
+__all__ = [
+    "fluent_astify",
+    "fluent_astify_message",
+    "fluent_parse",
+    "fluent_parse_message",
+    "fluent_serialize",
```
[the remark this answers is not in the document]
eemeli: Not really valid, because the result of `fluent_astify_message` is a Fluent pattern, and there is no universally valid canonical serialization of a Fluent pattern, given that the indent level of multiline content varies between message values and attributes, not to mention the selector values where Fluent patterns also show up.

moz/l10n/inc/parse.py
```diff
@@ -0,0 +1,55 @@
...
+# limitations under the License.
+
+from re import compile
+
+from ..resource import Comment, Entry, Resource, Section
+
+re_define = compile(r"#define[ \t]+(\w+)(?:[ \t](.*))?")
+
+
+def inc_parse(source: str) -> Resource[str, None]:
```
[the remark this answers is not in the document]
eemeli: It's not. Silme's inc parser matches its other format parsers in structure, and it doesn't take into account the "linedness" of the format, so this way is simpler.

moz/l10n/po/parse.py
```diff
@@ -0,0 +1,67 @@
...
+- "flags": `list[str]`
+- "plural": `str`
+"""
+
+
+def po_parse(source: str) -> Resource[tuple[str, ...], PoMetadataType]:
+    """
+    Parse a .po or .pot file into a message resource
+    """
+    pf = pofile(source)
```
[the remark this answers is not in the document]
eemeli: Ah, looks like that's missing from the entry metadata. Will need to fix that.

moz/l10n/properties/serialize.py
eemeli: Identifiers are lists of strings, so something like `['foo', 'bar']`. Parsing .properties will always result in identifiers with length 1, but if you then post-process the resource to e.g. split an identifier like `['foo.bar']` into `['foo', 'bar']` you'll get some. They also show up in resources parsed from Fluent, where we have ids like `['name', 'attr']`.

moz/l10n/properties/serialize.py
eemeli: Not serialize content like section identifiers which isn't natively supported by .properties, or define a transform on the parsed resource that's aware of your peculiarities.

tests/data/test.properties
eemeli: With this library, leading and trailing whitespace in .properties values is properly escaped: https://github.com/mozilla/moz-l10n/blob/4acc215109e02b560f8a0a1d031db1689e522b84/moz/l10n/properties/serialize.py#L93-L96

tests/test_fluent.py
```diff
@@ -0,0 +1,411 @@
...
+        self.assertEqual(res1, res2)
+
+    def test_equality_different_whitespace(self):
+        source1 = b"foo = { $arg }"
+        source2 = b"foo = {    $arg    }"
+        res1 = fluent_parse(source1, fluent_parse_message)
+        res2 = fluent_parse(source2, fluent_parse_message)
+        self.assertEqual(res1, res2)
+
+    def test_resource(self):
```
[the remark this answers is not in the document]
eemeli: I inlined the FTL contents because `FluentParser.parse()` requires a `str`, and so there was no reason to test its `bytes` support as with some of the other formats.

Adding the `demo.ftl` here does seem like a good idea.

tests/test_inc.py
```diff
@@ -0,0 +1,90 @@
...
+        self.assertEqual(
+            res,
+            Resource(
+                [
+                    Section(
+                        [],
+                        [
+                            Comment("#filter emptyLines"),
+                            Entry(["MOZ_LANGPACK_CREATOR"], "SeaMonkey e.V."),
+                            Comment(
```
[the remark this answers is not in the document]
eemeli: I think we'll need to keep that in Pontoon. The .inc format is already pretty special, and considering that key to have a special meaning in the syntax would be... special.

README.md
```diff
@@ -0,0 +1,20 @@
...
+and `FORMAT_serialize(resource)` will always provide a `str` iterator.
```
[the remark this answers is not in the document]
eemeli: I renamed the "source" as "text". Kinda hoping that "resource" connects well enough with the "Resource" mentioned in the preceding paragraph? Also, this readme will obviously need to be expanded later once we've a more solid set of features.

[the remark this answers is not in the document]
mathjazz: I think an example would help.

tests/test_po.py
```diff
@@ -0,0 +1,210 @@
...
+from unittest import TestCase
+
+from moz.l10n.po import po_parse, po_serialize
+from moz.l10n.resource import Entry, Metadata, Resource, Section
+
+# Show full diff in self.assertEqual. https://stackoverflow.com/a/61345284
+# __import__("sys").modules["unittest.util"]._MAX_LENGTH = 999999999
+
+
+class TestPo(TestCase):
```
mathjazz, line 26: This file would be easier to read if assertions would be split into separate methods.

A comment explaining what each assertion does would also help.

tests/test_po.py
```diff
@@ -0,0 +1,210 @@
...
+msgstr[2] "%d prevedeni sporočili"
+msgstr[3] "%d prevedena sporočila"
+
+#~ msgid "obsolete string"
+#~ msgstr "translated string"
+
+msgid "other string"
+msgstr "translated string"
+
+msgctxt "context"
```
mathjazz, line 126: It's a shame that the strings with context now appear at the bottom of the file.

tests/test_fluent.py
```diff
@@ -409,3 +410,369 @@ def test_meta(self):
...
+            ),
+        ]
+        self.assertEqual(res, Resource(sections=[Section(id=[], entries=entries)]))
+        self.assertEqual(
+            "".join(fluent_serialize(res)),
+            dedent(
+                """\
+                # Simple string
+                title = About Localization
+                # Multiline string: press Shift + Enter to insert new line
```
mathjazz, line 669: Can we preserve newlines?

I worry about the diff size.

eemeli: If you mean newlines e.g. before this comment, then no, we can't. They are not relevant for the resource contents, and therefore are left out of the data model.

And actually, doing so (and otherwise normalising the syntaxes) will minimize diff sizes. The key thing to keep in mind here is that in practice we don't have anywhere where both humans _and_ automation need to both manage contents.

tests/test_po.py
eemeli: Yes, but that's effectively how the structure works within .po files, given that the context is shared between multiple messages and how it's a part of what identifies each.

tests/test_po.py
eemeli: Split done.

tests/test_properties.py
```diff
@@ -0,0 +1,269 @@
...
+from importlib.resources import files
+from textwrap import dedent
+from unittest import TestCase
+
+from moz.l10n.properties import properties_parse, properties_serialize
+from moz.l10n.resource import Entry, Resource, Section
+
+
+class TestProperties(TestCase):
+    def testBackslashes(self):
```
mathjazz, line 24: Is there are reason not to use the snake_case here and in the next test?

tests/test_fluent.py
mathjazz: SGTM. Let's keep PMs in the loop about this.

mathjazz (approved):
Nice job! Looking forward to it being used in the wild.

tests/test_properties.py
eemeli: No real reason, this is a remnant from having been adopted from [here](https://github.com/mozilla/compare-locales/blob/5f2e98b587d1e2193b4531aea5890e3aa510aacf/compare_locales/parsers/tests/test_properties.py#L15), which used camelCase.

PR #2, opened 2024-04-11
Add support for web extension messages.json & plain key-value JSON files

eemeli:
The parser and serialiser are pretty strict for this, as the `messages.json` format is pretty explicit. It also effectively requires using the MF2 data model representation, does not allow for metadata, and is also pretty strict on the keys it supports.

### DIFF
--- a/moz/l10n/plain/__init__.py
+++ b/moz/l10n/plain/__init__.py
@@ -1,0 +1,4 @@
+from .parse import plain_parse
+from .serialize import plain_serialize
+
+__all__ = ["plain_parse", "plain_serialize"]
--- a/tests/test_plain.py
+++ b/tests/test_plain.py
@@ -1,0 +1,123 @@
+# Copyright Mozilla Foundation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from importlib.resources import files
+from textwrap import dedent
+from unittest import TestCase
+
+from moz.l10n.plain import plain_parse, plain_serialize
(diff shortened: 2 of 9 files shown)


eemeli:
Also added nested key-value JSON files.

tests/test_plain.py
```diff
@@ -0,0 +1,123 @@
...
+                                "Hello$$$ $1YOUR_NAME$ at $2",
+                            ),
+                            Entry(["placeholders", "description"], "Peer greeting"),
+                            Entry(
+                                [
+                                    "placeholders",
+                                    "placeholders",
+                                    "1your_name",
+                                    "content",
+                                ],
```
mathjazz, line 62: Interesting, I didn't even realize Pontoon already supports multiple nesting levels in key-value JSON files. Just checked and... it does.

tests/test_webext.py
```diff
@@ -0,0 +1,171 @@
...
+                            ),
+                            Entry(
+                                ["NoCommentsorSources"],
+                                PatternMessage(["Translated No Comments or Sources"]),
+                            ),
+                            Entry(
+                                ["placeholders"],
+                                PatternMessage(
+                                    [
+                                        "Hello$$ ",
```
mathjazz, line 57: It puzzles me why is this not `"Hello$$$ "`.

mathjazz (approved):
I'd change the paths, other than that is looks good.

moz/l10n/plain/__init__.py
```diff
@@ -0,0 +1,4 @@
+from .parse import plain_parse
+from .serialize import plain_serialize
+
+__all__ = ["plain_parse", "plain_serialize"]
```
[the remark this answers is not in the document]
eemeli: That was an intentional choice, to account for the possibility of this structure working with other resource formats, such as YAML.

tests/test_webext.py
eemeli: From [here](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/i18n/Locale-Specific_Message_reference#message):
> Any number of consecutive dollar signs appearing in strings are replaced by the same number of dollar signs minus one. So, $$ > $, $$$ > $$, etc.

eemeli:
I'm going to keep the paths for now, and iterate on them later as the whole shape of this comes into clearer focus.

We could e.g. end up with `moz.l10n.resource.webext`, and going that deep adding a `_json` there starts to feel like one more hierarchy layer, even if it's just in the name.

PR #3, opened 2024-04-14
Add Android string resource support

eemeli:
Handles all parts of the resource format, except for ICU MessageFormat strings.

### DIFF
--- a/moz/l10n/android/parse.py
+++ b/moz/l10n/android/parse.py
@@ -1,0 +1,332 @@
+# Copyright Mozilla Foundation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from collections.abc import Callable, Iterable, Iterator
+from re import DOTALL, compile, fullmatch, sub
+from typing import cast
+
+from lxml import etree
+
+from ..message import (
+    CatchallKey,
+    Expression,
+    FunctionAnnotation,
+    Markup,
+    Message,
+    PatternMessage,
+    SelectMessage,
+    VariableRef,
+)
+from ..resource import Comment, Entry, Metadata, Resource, Section
+
+plural_categories = ("zero", "one", "two", "few", "many", "other")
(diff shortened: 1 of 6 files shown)


tests/data/strings.xml
```diff
@@ -0,0 +1,54 @@
...
+  <!-- standalone -->
+
+  <string name="welcome">Welcome to <b>&foo;</b>!</string>
+  <string name="placeholders">Hello, %1$s! You have %2$d new messages.</string>
+  <string name="escape_html">Hello, %1$s! You have &lt;b>%2$d new messages&lt;/b>.</string>
+
+  <string name="ws_trimmed"> &#32; &#8200; &#8195;</string>
+  <string name="ws_quoted">" &#32; &#8200; &#8195;&quot;</string>
+  <string name="ws_escaped"> \u0032 \u8200 \u8195</string>
+
```
mathjazz, line 25: I would also expect apostrophes to be tested and handled properly:
https://github.com/mozilla/pontoon/pull/3020/files#diff-e8bac283d0aa4db32a8a4a7f254bc66cadeeb611975a0d1beb8d5ecb4979dc80R324

tests/data/strings.xml
```diff
@@ -0,0 +1,54 @@
...
+    <item>Mars</item>
+  </string-array>
+
+  <plurals name="numberOfSongsAvailable">
+    <!--
+         As a developer, you should always supply "one" and "other"
+         strings. Your translators will know which strings are actually
+         needed for their language. Always include %d in "one" because
+         translators will need to use %d for languages where "one"
+         doesn't mean 1 (as explained above).
```
mathjazz, line 44: Where "above"?

mathjazz (commented):
This is impressive. Android plurals!

tests/data/strings.xml
eemeli: Ah, a remnant from the [Android docs](https://developer.android.com/guide/topics/resources/string-resource.html#Plurals) from where I lifted this.

mathjazz (approved):

eemeli:
Alas, needed to refactor this still a bunch to properly support `<xliff:g>`, which can be used to mark parts of the message as nontranslatable.

Also, it turns out that at least some of the Android escaping is valid within nested elements. But not all of it.

Please take another look?

mathjazz (approved):
The code looks good, although I found it a bit tricky to review, so I'm mostly relied on the tests. I'm confident we'll spot any issues as part of integrating this code into Pontoon and testing with the real files.

I'd only add that link to the <xliff:g> reference.

moz/l10n/android/parse.py
```diff
@@ -0,0 +1,414 @@
...
+#   - Strings with ICU MessageFormat contents.
+#     These also use "quotes" for special escaping behaviour.
+#     ICU MessageFormat strings are not currently detected by this library.
+#
+# The source contents of each of the above needs to be parsed differently,
+# and message strings can be found in <string>, <string-array>, and <plurals>
+# elements, each of which also needs different parsing.
+#
+# For more information, see:
+# https://developer.android.com/guide/topics/resources/string-resource
```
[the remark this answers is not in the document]
eemeli:
```suggestion
# https://developer.android.com/guide/topics/resources/string-resource
# https://developer.android.com/guide/topics/resources/localization#mark-message-parts
```

PR #5, opened 2024-04-29
Add generic parse/serialize methods, making format handlers more self-similar

eemeli:
Overall, improves alignment between different formats, simplifying their APIs and ensuring that all can support resources with `str` or `Message` messages.

Adds a format detector, based on the file name and where necessary, the file contents.

Based on the above, adds two top-level functions to `moz.l10n.resource`:

```python
def parse_resource(
    type: Format | str | None, source: str | bytes
) -> Resource[Message, str]: ...

def serialize_resource(
    resource: Resource[str, str] | Resource[Message, str],
    format: Format | None = None,
    trim_comments: bool = False,
) -> Iterator[str]: ...
```

Moves individual resource handlers under `moz.l10n.resource`, and renames `plain` as `plain_json`.


eemeli:
Updated the README from docstrings, and added a new utility function:

```python
def iter_resources(
    root: str,
    dirs: list[str] | None = None,
    ignorepath: str = ".l10n-ignore"
) -> Iterator[tuple[str, Resource[Message, str] | None]]
```

> Iterate through localizable resources under the `root` directory. Use `dirs` to limit the search to only some subdirectories under `root`.
>
> Yields `(str, Resource | None)` tuples, with the file path and the corresponding `Resource`, or `None` for files that could not be parsed as localization resources.
>
> To ignore files, include a `.l10n-ignore` file in `root`, or some other location passed in as `ignorepath`. This file uses a git-ignore syntax, and is always based in the `root` directory.

moz/l10n/resource/parse_resource.py
```diff
@@ -0,0 +1,63 @@
...
+from .ini.parse import ini_parse
+from .plain_json.parse import plain_json_parse
+from .po.parse import po_parse
+from .properties.parse import properties_parse
+from .webext.parse import webext_parse
+from .xliff.parse import xliff_parse
+
+
+def parse_resource(
+    type: Format | str | None, source: str | bytes
```
mathjazz, line 31: Why is the `type` parameter required?

I would expect the file path to be often passed to the function implicitly through the `source` parameter. It seems redundant to then be required to also include `type=None` as the function parameter.

tests/test_parse_serialize_resource.py
```diff
@@ -0,0 +1,74 @@
...
+            "messages.json",
+            "strings.xml",
+            "test.properties",
+            "xcode.xliff",
+        )
+        for file in data:
+            res = parse_resource(file, get_source(file))
+            assert isinstance(res, Resource)
+            list(serialize_resource(res))
+            list(serialize_resource(res, trim_comments=True))
```
mathjazz, line 46: What's the purpose of the last two lines without an `assert`?

mathjazz (commented):
Nice work!

Minor comments inline only.

README.md
```diff
@@ -19,3 +19,71 @@ but additional input types and options vary by format.
...
+def serialize_resource(
+    resource: Resource[str, str] | Resource[Message, str],
+    format: Format | None = None,
+    trim_comments: bool = False
+) -> Iterator[str]
+```
+
+Serialize a Resource as its string representation.
+
+If `format` is set, it overrides the `resource.format` value.
```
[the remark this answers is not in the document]
eemeli: This lets you convert content from one format to another, or to generate a resource from scratch. Allowing for this possibility ensures that the individual serialisers don't just assume that their input is exactly in the shape that their parsers emit, but actually check their input.

Specifically allowing for this separately from the `resource.format` value allows for it to be kept immutable.

moz/l10n/resource/__init__.py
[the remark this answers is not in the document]
eemeli: That doesn't quite work, because I'm pretty sure that eventually the "message" part will need to be expanded to support MF1, MF2, and probably other message formats as well -- possibly including some generalisation of printf.

Hence the current division into "resource formats" and "message formats".

Happy to continue iterating on the structure some more.

moz/l10n/resource/parse_resource.py
eemeli: Because `source` holds the actual contents of the file, it won't have the file path. So you can pass the path in with `type: str`, as in `parse_resource('/foo/bar.json', source)`.

tests/test_parse_serialize_resource.py
eemeli: They exhaust the generator. This ensures that the complete resource can be serialized without error, and not just the first line. Is there a more pythonic way of doing that?

moz/l10n/resource/__init__.py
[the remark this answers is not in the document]
mathjazz: > Hence the current division into "resource formats" and "message formats".

I don't see that division (as least in the folder structure).

> Happy to continue iterating on the structure some more.

OK, as long as this is not set in stone, I'm ok with it. Also, it's a nit.

moz/l10n/resource/parse_resource.py
mathjazz: > So you can pass the path in with `type: str`, as in `parse_resource('/foo/bar.json', source)`.

And that's the thing that feels redundant - `parse_resource('/foo/bar.json')` should be sufficient.

tests/test_parse_serialize_resource.py
mathjazz: I see. Something like https://stackoverflow.com/a/42327075 perhaps? A comment would also work.

mathjazz (approved):
No more blocking comments on my end. I'd still at least add a comment on that test.

moz/l10n/resource/parse_resource.py
eemeli: That makes sense. I made the second argument optional, if the first one is a string path.

PR #6, opened 2024-05-16
Add automagical & configurable resource path iterators

eemeli:
This implements functionality currently provided by [`compare_locales.paths`](https://github.com/mozilla/compare-locales/tree/release/compare_locales/paths), [`pontoon.sync.vcs`](https://github.com/mozilla/pontoon/tree/main/pontoon/sync/vcs), and [`pontoon.sync.utils`](https://github.com/mozilla/pontoon/blob/main/pontoon/sync/utils.py).

Similar APIs are provided for the configured and non-configured cases.

The `iter_resources()` is generalised into `walk_files()`, dropping its internal `parse_resource()` calls.

Making us of the paths, a command-line tool `l10n-fix` is included to fix the formatting of supported localization files.

As a bit of feature creep, a fuzzy resource comparator `moz.l10n.resource.l10n_equal()` is also added, and entry & sections identifiers are made immutable.


moz/l10n/paths/discover.py
```diff
@@ -0,0 +1,235 @@
...
+        self,
+        root: str,
+        ref_root: str | None = None,
+        ignorepath: str | None = ".l10n-ignore",
+    ) -> None:
+        """
+        To ignore files, include a `.l10n-ignore` file in the reference base directory
+        or some other location passed in as `ignorepath`.
+        This file uses git-ignore syntax,
+        and is always based in the reference base directory.
```
mathjazz, line 86: What does _based in_ mean? Located in?

tests/test_discover.py
```diff
@@ -0,0 +1,195 @@
...
+                "two/a.ftl": "two/a.ftl",
+                "two/b.pot": "two/b.po",
+            }.items()
+        }
+        assert paths.all() == expected
+        for ref, tgt in expected:
+            assert paths.target_path(ref) == tgt
+            assert paths.target_path(ref, locale="xx") == tgt.format(locale="xx")
+
+    def test_ref_priorities(self):
```
mathjazz, line 94: To actually test priority order, multiple reference directories should co-exist.

mathjazz (commented):
It would be nice to add moz.l10n.paths to the README.

Other than that I don't have major concerns.

Can't wait to test this in the wild.

moz/l10n/paths/config.py
```diff
@@ -0,0 +1,294 @@
...
+    Wrapper for localization config files.
+
+    Supports a subset of the format specified at:
+    https://moz-l10n-config.readthedocs.io/en/latest/fileformat.html
+
+    Differences:
+    - `[build]` is ignored
+    - `[[excludes]]` are not supported
+    - `[[filters]]` are ignored
+    - `[[paths]]` must always include both `reference` and `l10n`
```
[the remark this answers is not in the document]
eemeli: With some conditions, yes.

- `[build]` is completely unused atm.
- There is [one](https://github.com/mozilla-l10n/www-l10n/blob/cd05d22d972eaa802475a1503a7ff4d7be0acd87/l10n-pontoon.toml#L10) `[[excludes]]` value we'll need to account for somehow.
- The SeaMonkey and Thunderbird `[[filters]]` will need some review; the [one](https://github.com/mozilla/gecko-dev/blob/9fcc11127fbfbdc88cbf37489dac90542e141c77/browser/locales/l10n.toml#L164) we still have for Firefox should be safe to ignore.
- I checked; all `[[paths]]` values include both fields.

moz/l10n/paths/discover.py
eemeli: Means that it's the root directory for the expressions, even if the `.l10n-ignore` file is itself at a different path.

tests/test_discover.py
eemeli: They do co-exist here; each of the `build_file_tree()` commands is cumulative on the previous, as they're done within the same `with TemporaryDirectory()` block.

eemeli:
> It would be nice to add `moz.l10n.paths` to the README.

Done.

moz/l10n/paths/config.py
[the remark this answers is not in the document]
eemeli: To clarify: This isn't ready for all use cases yet, but it's a step towards being able to identify exactly what of the remainder we'll actually need.

mathjazz (approved):

PR #9, opened 2024-07-04
Fix more small bugs

eemeli:
This is a grab-bag of small stuff, discovered by testing the tools against [firefox-l10n](https://github.com/mozilla-l10n/firefox-l10n) contents.

The test changes replacing `self.assertEqual()` with `assert ...` are because they result in clearer error logs with `pytest`, whereas the former was better with `python -m unittest`.


mathjazz (approved):
LGTM

PR #12, opened 2024-07-10
properties: Improve character escape handling

eemeli:
This accounts for the upstream fix for translate/translate#5199.


mathjazz (approved):
LGTM

PR #14, opened 2024-07-11
Add l10n-build command-line tool

### LINKED ISSUE #13: Add support for filtering locale data
Currently, Firefox localization works by first collecting in [firefox-l10n-source](https://github.com/mozilla-l10n/firefox-l10n-source) the localizable contents of all active branches, and by localizing them in [firefox-l10n](https://github.com/mozilla-l10n/firefox-l10n). During the Firefox build, these localized files are used directly for both repacks and langpacks, such that contents for all branches' localizations are included in the assets of each branch.

This ultimately leads to issues such as [bug 1899255](https://bugzilla.mozilla.org/show_bug.cgi?id=1899255) when we don't remember to account for all the details of how changes affect other branches.

To make the behaviour of localization more predictable and less prone to causing failures, we should build tooling to filter each localization's contents to those actually used in a given branch, and to ensure that all files exist, even if they're empty.

This will also allow for filtering out localization notes and other comments from the files, as they're not needed for formatting.

eemeli:
Fixes #13 
CC @bhearsum

The idea would be for this tool to be run during the Firefox build, at which time it would trim out all the contents from localization files that are not intended for the current branch.

The filtering can be done in-place (with `--base` and `--target` the same path), but then it won't delete files that are in the localization that are not in the reference source. The target directory is not clobbered.

On my M1 Macbook, this takes about 3s per standalone locale, and about 130s when run on all locales, using the `browser/locales/l10n.toml` config. This is probably fast enough that the intermediate results don't need to be cached e.g. in `firefox-l10n` branches, and can be produced as a part of the build.

An initial test would indicate that this filtering will reduce compressed langpack sizes by about a third.

### DIFF
--- a/moz/l10n/bin/build.py
+++ b/moz/l10n/bin/build.py
@@ -117,7 +117,7 @@ def build_targets_for_release(
                 else:
                     log.info(f"skip {rel_path}")
             elif source_ids:
-                log.info(f"create {rel_path}")
+                log.info(f"create empty {rel_path}")
                 open(tgt_path, "a").close()
             else:
                 log.info(f"skip {rel_path}")
--- a/moz/l10n/bin/build.py
+++ b/moz/l10n/bin/build.py
@@ -1,0 +1,127 @@
+# Copyright Mozilla Foundation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import annotations
+
+import logging
+from argparse import ArgumentParser, RawDescriptionHelpFormatter
+from os import makedirs
+from os.path import dirname, exists, join, relpath
+from shutil import copyfile
+from textwrap import dedent
+
+from moz.l10n.paths.config import L10nConfigPaths
+from moz.l10n.resource import UnsupportedResource, parse_resource, serialize_resource
+from moz.l10n.resource.data import Entry
+
+log = logging.getLogger(__name__)
+
+
+def cli() -> None:
+    parser = ArgumentParser(
+        description=dedent(
+            """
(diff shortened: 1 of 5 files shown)


bhearsum:
> An initial test would indicate that this filtering will reduce compressed langpack sizes by about a third.

That's a substantial decrease!

moz/l10n/bin/build.py
```diff
@@ -0,0 +1,127 @@
...
+            if exists(l10n_path):
+                if source_ids:
+                    msg_delta = 0
+                    res = parse_resource(l10n_path)
+                    for section in res.sections:
+                        msg_delta -= len(section.entries)
+                        section.entries = [
+                            entry
+                            for entry in section.entries
+                            if not isinstance(entry, Entry)
```
flodolo, line 105: Why do we need this check? What can be a "non Entry" within a section of a localized file?

moz/l10n/bin/build.py
```diff
@@ -0,0 +1,127 @@
...
+                    with open(tgt_path, "w") as file:
+                        for line in serialize_resource(res, trim_comments=True):
+                            file.write(line)
+                elif l10n_base != l10n_target:
+                    log.info(f"copy {rel_path}")
+                    copyfile(l10n_path, tgt_path)
+                else:
+                    log.info(f"skip {rel_path}")
+            elif source_ids:
+                log.info(f"create {rel_path}")
```
flodolo, line 120:
```suggestion
                log.info(f"create empty {rel_path}")
```

Might be clearer.

flodolo (approved):
This trims away license headers. Are we OK with that, or should they be treated differently from other comments? 

I think this would benefit from @mathjazz review, since he's more familiar with the underlying data structure at this point.

moz/l10n/bin/build.py
eemeli: Sections may contain standalone comments, which are `Comment` instances.

eemeli:
> This trims away license headers. Are we OK with that, or should they be treated differently from other comments?

This is intentional, and fine. The output of the build tool is not the preferred form for making modifications, so it's not considered as the "Source" or "Source Code Form" but as an "Object", "Executable Form", or "binary form" of the work as understood by e.g. MPL2, Apache 2.0, or BSD licenses. The format of the resulting file is not relevant for these considerations. This means that the license header does not need to be retained.

mathjazz (approved):
I didn't actually test the changes, but they look good.

Nice work downsizing the langpacks!

bcolsson (approved):

PR #19, opened 2024-08-06
Expand use_source_values to use_source_entries in add_entries()

eemeli:
Context: mozilla-l10n/firefox-l10n-source#73

### DIFF
--- a/tests/test_add_entries.py
+++ b/tests/test_add_entries.py
@@ -40,9 +40,15 @@ class TestAddEntries(TestCase):
     def test_message_changed_in_source_use_source_values(self):
         target = Resource(None, [Section((), [Entry(("foo",), "Foo 1")])])
         source = Resource(None, [Section((), [Entry(("foo",), "Foo 2")])])
-        self.assertEqual(add_entries(target, source, use_source_values=True), 1)
-        self.assertEqual(
-            target, Resource(None, [Section((), [Entry(("foo",), "Foo 2")])])
+        assert add_entries(target, source, use_source_entries=True) == 1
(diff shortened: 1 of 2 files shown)


tests/test_add_entries.py
```diff
@@ -40,9 +40,15 @@ def test_message_changed_in_source(self):
...
         target = Resource(None, [Section((), [Entry(("foo",), "Foo 1")])])
         source = Resource(None, [Section((), [Entry(("foo",), "Foo 2")])])
-        self.assertEqual(add_entries(target, source, use_source_values=True), 1)
-        self.assertEqual(
-            target, Resource(None, [Section((), [Entry(("foo",), "Foo 2")])])
+        assert add_entries(target, source, use_source_entries=True) == 1
+        assert target == Resource(None, [Section((), [Entry(("foo",), "Foo 2")])])
+
+    def test_message_comment_changed_in_source_use_source_values(self):
+        target = Resource(None, [Section((), [Entry(("foo",), "Foo", "Bar 1")])])
```
flodolo, line 47: Is `Bar 1/2` the comment here? Could we use text that helps to understand the test without chasing down code definitions (e.g. `id1`, `value1`, `Comment 1`)?

eemeli: > Is `Bar 1/2` the comment here?

Yes.

> Could we use text that helps to understand the test without chasing down code definitions (e.g. `id1`, `value1`, `Comment 1`)?

Certainly. I mostly look at these in VS Code, where the type hinting provides that info on hover. Would be happy to apply the string changes as a separate follow-up PR.

flodolo (approved):

PR #21, opened 2024-09-06
Reduce minimum external dependencies and add line positions

eemeli:
The `translate-toolkit` and `lxml` dependencies form the vast majority of the filesize footprint of the library. To reduce that:
1. Write our own .properties parser and serialiser.
2. Treat the support of XML formats (Android, XLIFF) as an optional extra, installable as `moz.l10n[xml]`.

### DIFF
--- a/moz/l10n/resource/data.py
+++ b/moz/l10n/resource/data.py
@@ -39,6 +39,34 @@ The Message value type.
 
 
 @dataclass
+class LinePos:
+    """
+    The source line position of an entry or section header.
+    """
+
+    start: int
+    """
+    The starting line of the entry or section.
+    May be less than `value` if preceded by a comment.
+    """
+
+    key: int
+    """
+    The start line of the entry or section header key or name.
+    """
+
+    value: int
+    """
+    The start line of the entry pattern or section header.
+    """
+
+    end: int
+    """
+    The line one past the end of the entry or section header.
+    """
+
+
+@dataclass
 class Metadata(Generic[M]):
     """
     Metadata is attached to a resource, section, or a single entry.
@@ -122,6 +150,12 @@ class Entry(Generic[V, M]):
     Metadata attached to this entry.
     """
 
+    linepos: LinePos | None = None
+    """
+    The parsed position of the entry,
+    available for some formats.
+    """
+
 
 @dataclass
 class Section(Generic[V, M]):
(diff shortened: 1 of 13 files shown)


flodolo:
This is code that @mathjazz ought to review, especially given the Fluent related changes.

moz/l10n/resource/data.py
```diff
@@ -38,6 +38,34 @@
...
+    """
+
+    value: int
+    """
+    The start line of the entry pattern or section header.
+    """
+
+    end: int
+    """
+    The line one past the end of the entry or section header.
```
flodolo, line 65: Why "one past"? What happens at the end of a file if there's no empty line?

flodolo:
One more question: have you checked the output of the properties serializer against `firefox-l10n-source`? What would be the diff there?

moz/l10n/resource/data.py
eemeli: Then this'll point to a line that's not there. It's the same as with character ranges; the "end" pointer is not included in the range. This works even if there's no such character (or line) actually there.

eemeli:
> One more question: have you checked the output of the properties serializer against `firefox-l10n-source`? What would be the diff there?

The only difference is that for empty messages, the new serialiser does not include a space after the `=`. In `firefox-l10n-source` this affects the following files:
- devtools/client/toolbox.properties
- dom/chrome/xslt/xslt.properties
- toolkit/chrome/global-platform/mac/platformKeys.properties
- toolkit/chrome/global/dialog.properties
- toolkit/chrome/global/intl.properties

mathjazz (commented):
Changes look good and I trust the tests.

I'm curious about [Include line positions when parsing Fluent](https://github.com/mozilla/moz-l10n/pull/21/commits/471446937b799c81dd9793d65333b60eee645c38), since it's not mentioned in the description. Are line positions required for the new properties parser and serializer and you added them to Fluent as well?

mathjazz (approved):

flodolo:
> I'm curious about [Include line positions when parsing Fluent](https://github.com/mozilla/moz-l10n/pull/21/commits/471446937b799c81dd9793d65333b60eee645c38), since it's not mentioned in the description. Are line positions required for the new properties parser and serializer and you added them to Fluent as well?

The pull request title needs updating.

Having line positions is a requirement to use moz-l10n as a linter in `mozilla-central`.

flodolo (approved):

PR #22, opened 2024-09-16
Path improvements for use in Pontoon

eemeli:
A collection of fixes to paths required by Pontoon (mozilla/pontoon#3312).

### DIFF
--- a/moz/l10n/paths/discover.py
+++ b/moz/l10n/paths/discover.py
@@ -51,7 +51,7 @@ class L10nDiscoverPaths:
     """The target base directory, with subdirectories for each locale."""
     locales: list[str] | None
     """Locales detected from subdirectory names under `base`."""
-    ref_paths: tuple[str, ...]
+    ref_paths: list[str]
     """Reference paths"""
 
     def __init__(
@@ -59,6 +59,7 @@ class L10nDiscoverPaths:
         root: str,
         ref_root: str | None = None,
         *,
+        force_paths: list[str] | None = None,
         ignorepath: str | None = ".l10n-ignore",
         source_locale: str | list[str] | None = None,
     ) -> None:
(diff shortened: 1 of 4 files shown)


moz/l10n/paths/discover.py
```diff
@@ -78,9 +58,31 @@ def __init__(
...
+
+        The reference directory has a name matching the `source_locale`
+        and contains files with extensions that appear localizable.
+
+        If `ref_root` is given, the reference directory must be within it.
+        If it does not contain a directory matching `source_locale`
+        but `ref_root` itself contains files with extensions that appear localizable,
+        it is used as the reference directory.
+
+        The default `source_locale` is `['en-US', 'en']`,
```
mathjazz, line 83: What happens with "templates"?

eemeli: Not treated as anything special. From Pontoon we [call this](https://github.com/mozilla/pontoon/pull/3312/files#diff-b18de80c1f885538e633fd3935679aa1db868c06afc4eecf9af1341df7b707aaR29) with `ref_root` set to the source repo root, which doesn't need to have any special name in order to fall back to using the repo root as the reference root.

mathjazz: How is it going to detect that https://github.com/mozilla/addons-server/tree/master/locale/templates is the source locale folder, and not https://github.com/mozilla/addons-server/tree/master/locale/en_US?

eemeli: At least for now, we'll need to call this from Pontoon with `source_locale=["templates", "en-US", "en"]` to keep its behaviour consistent.

eemeli:
I think this batch is complete for now, and I could roll out a new release once this lands. Further iteration may of course be needed later.

moz/l10n/paths/discover.py
mathjazz: Meh, I completely missed `ref_paths` was a function parameter.

mathjazz (approved):

PR #24, opened 2024-10-03
Add l10n-build-file command, fill missing messages from source

eemeli:
Three related improvements:
1. Adds a CLI command `l10n-build-one`, to replace [l10n_merge.py](https://searchfox.org/mozilla-central/source/python/mozbuild/mozbuild/action/l10n_merge.py) that's used by the m-c FasterMake backend.
2. In `l10n-build` and `l10n-build-one`, ensure that non-Fluent target files are complete by using the source as fallback.
3. Adds a per-locale summary to the `l10n-build` output.

### DIFF
--- a/moz/l10n/bin/build.py
+++ b/moz/l10n/bin/build.py
@@ -66,9 +66,7 @@ def cli() -> None:
     log_level = (
         logging.WARNING
         if args.verbose == 0
-        else logging.INFO
-        if args.verbose == 1
-        else logging.DEBUG
+        else logging.INFO if args.verbose == 1 else logging.DEBUG
     )
     logging.basicConfig(format="%(message)s", level=log_level)
 
--- a/moz/l10n/bin/build.py
+++ b/moz/l10n/bin/build.py
@@ -63,64 +63,78 @@ def cli() -> None:
     log_level = (
         logging.WARNING
         if args.verbose == 0
-        else logging.INFO if args.verbose == 1 else logging.DEBUG
+        else logging.INFO
+        if args.verbose == 1
+        else logging.DEBUG
     )
     logging.basicConfig(format="%(message)s", level=log_level)
 
-    build_targets_for_release(args.config, args.base, args.target, set(args.locales))
+    cfg_path: str = args.config
+    l10n_base: str = args.base
+    l10n_target: str = args.target
+    locales: set[str] = set(args.locales)
 
-
-def build_targets_for_release(
-    cfg_path: str, l10n_base: str, l10n_target: str, locales: set[str]
-) -> None:
     paths = L10nConfigPaths(cfg_path)
     paths.base = l10n_base
     paths.locales = None
     for (source_path, l10n_path_template), path_locales in paths.all().items():
         log.debug(f"source {source_path}")
-        try:
-            source = parse_resource(source_path)
-            source_ids = set(
-                section.id + entry.id
-                for section in source.sections
-                for entry in section.entries
-                if isinstance(entry, Entry)
-            )
-        except UnsupportedResource:
-            source_ids = None
+        source_ids = get_source_message_ids(source_path)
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -42,6 +42,7 @@ xml = ["lxml ~= 5.0"]
 
 [project.scripts]
 l10n-build = "moz.l10n.bin.build:cli"
+l10n-build-one = "moz.l10n.bin.build_one:cli"
```
flodolo, line 45: Maybe `single` instead of `one`? For some reason, the latter sounds strange to me.

moz/l10n/bin/build_one.py
```diff
@@ -0,0 +1,69 @@
...
+
+    log_level = (
+        logging.WARNING
+        if args.verbose == 0
+        else logging.INFO if args.verbose == 1 else logging.DEBUG
+    )
+    logging.basicConfig(format="%(message)s", level=log_level)
+
+    try:
+        source_res = parse_resource(args.source)
```
flodolo, line 60: This fails badly if the resource is not a file.

moz/l10n/bin/build_one.py
```diff
@@ -0,0 +1,69 @@
...
+        makedirs(dirname(args.target), exist_ok=True)
+        write_target_file(args.source, source_res, args.l10n, args.target)
```
flodolo, line 62: And this fails badly if the target is not a file name (e.g. it's a folder).

moz/l10n/bin/build.py
```diff
@@ -67,61 +70,144 @@ def cli() -> None:
...
+            section.entries = [
+                entry
+                for entry in section.entries
+                if isinstance(entry, Entry) and section.id + entry.id in source_map
+            ]
+            msg_delta += len(section.entries)
+        msg = f"filter {name}"
+        log.info(f"{msg} ({msg_delta})" if msg_delta != 0 else msg)
+        with open(tgt_path, "w") as file:
+            for line in serialize_resource(l10n_res, trim_comments=True):
```
flodolo, line 154: I think you lost me here: you're trimming comments here, but where are you removing obsolete strings from the localized resource?

`l10n_res` hasn't been changed since `parse_resource(l10n_path) if exists(l10n_path) else None` (if the l10n resources is not empty).

flodolo:
Also, going back to our conversation on priorities, it's worth noting that this adds `NUMBER()` to localized files, so we'd need to investigate that issue before using this in production builds.

moz/l10n/bin/build_one.py
eemeli: That's kind of intentional? What else should happen if you try this with an invalid path, except getting something like a `FileNotFoundError` or an `IsADirectoryError`?

moz/l10n/bin/build.py
eemeli: The filtering of `l10n_res` contents was done here in the assignment to `section.entries` on line 145, but I refactored the Fluent handling to also be based on iterating the source, so this is out of date.

moz/l10n/bin/build_one.py
flodolo: Exit with an error without a full stack trace seems like a good solution?

eemeli:
> Also, going back to our conversation on priorities, it's worth noting that this adds `NUMBER()` to localized files, so we'd need to investigate that issue before using this in production builds.

I'll address that in a separate PR.

moz/l10n/bin/build.py
flodolo: Completely missed that `for section in l10n_res.sections` would affect `l10n_res` as an object 🤦🏼

flodolo (approved):
See note on exiting with an error instead of crashing, but otherwise this looks good (and a lot simpler)

PR #26, opened 2024-10-03
fluent: Strip bare NUMBER() from selectors when serialising

eemeli:
This circumvents the issues discovered in [bug 1918729](https://bugzilla.mozilla.org/show_bug.cgi?id=1918729) by removing the `NUMBER()` wrapper from selectors when serialising Fluent.


flodolo (approved):
Side note: on top of en-US, we'll need to clean up also locales at this point, and PDF.js :-( (otherwise we'll have the linter failing again).

PR #27, opened 2024-10-07
Use uv as package manager

eemeli:
This allows for the dev dependencies to be listed in `pyproject.toml`.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -17,25 +17,30 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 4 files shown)


flodolo (approved):
We should drop 3.8 at some point, since it's reaching [EOL this month](https://peps.python.org/pep-0569/).

eemeli:
Given that we want to use moz.l10n in m-c, the minimum supported version is determined by [`mach`](https://searchfox.org/mozilla-central/rev/b546e74055186bb387ed7db9079374e1e7e0b88e/mach#13).

PR #28, opened 2024-10-11
Improve config paths support

eemeli:
1. Support more wildcard patterns, in particular ones that include a `*` or `**` before a `{locale}` or other named variable in the L10n path.
2. Fix target path matching for L10n paths that start with `{l10n_base}`.

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -22,7 +22,7 @@ from textwrap import dedent
 from typing import Any, Dict, Union
 from unittest import TestCase
 
-from moz.l10n.paths import L10nConfigPaths
+from moz.l10n.paths import L10nConfigPaths, get_android_locale
 
 if sys.version_info >= (3, 11):
     from tomllib import load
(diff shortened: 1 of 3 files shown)


tests/test_config.py
```diff
@@ -333,3 +337,65 @@ def test_fenix(self):
...
+        assert paths.base == root
+        assert paths.locales is None
+        res_path = normpath(
+            "mozilla-mobile/android-components/components/foo/src/main/res"
+        )
+        exp_ref = join(root, res_path, normpath("values/strings.xml"))
+        exp_tgt = join(root, res_path, normpath("values-{android_locale}/strings.xml"))
+        assert paths.all() == {(exp_ref, exp_tgt): None}
+        assert paths.target(exp_ref) == (exp_tgt, ())
+        assert paths.find_reference("values-xx/strings.xml") is None
```
flodolo, line 393: Worth adding a test for a locale like en-GB?

flodolo (approved):

PR #30, opened 2024-11-20
l10n-compare: Conflate anonymous sections together

eemeli:
This should resolve some of the comparison issues seen in `firefox-l10n-source`.


bcolsson (approved):
This makes sense - can you double check my understanding: 

```
id0 = translation

## Localization Section
id1 = translation
id2 = translation

##

id3 = translation
```
In this case id0 and id3 would be combined because they're both anonymous sections. Or does the closing group comment create a second section?

eemeli:
> In this case id0 and id3 would be combined because they're both anonymous sections.

Yes.

PR #34, opened 2024-11-26
l10n-build: Copy unsupported files to target

eemeli:
Same as #33, but for the other build command, which I mis-remembered already working this way.


mathjazz (approved):
LGTM

PR #38, opened 2025-01-20
Move some files around

eemeli:
This is preparation and cleanup for adding MF2 support as a next step, which highlights that not all formats are resource-based.

Some of the nominally public paths change here (e.g. `moz.l10n.resource.fluent` -> `moz.l10n.formats.fluent`), but as far as I can tell, none of these have any external users.

Docs are updated to match.


mathjazz (approved):
LGTM

PR #39, opened 2025-01-20
Update & simplify message data model

eemeli:
This builds on #38, and will need a rebase after that's merged. Actual diff here is [`reorg-paths...simpler-messages`](https://github.com/mozilla/moz-l10n/compare/reorg-paths...simpler-messages).

This applies the following upstream changes made to the MF2 data model:
- unicode-org/message-format-wg#845
  - Do not allow variable references as attribute values
- unicode-org/message-format-wg#877
  - Use variables rather than expressions as selectors
- unicode-org/message-format-wg#883
  - "unsupported" constructs are dropped
  - `Expression.annotation` is renamed as `Expression.function`

Going further from there, the message data model is also made more ergonomic:
- Declarations are required for `SelectMessage`, as at least one is needed for the selector.
- Declarations are defined by a dict, rather than a list of `(name, expression)` tuples.
- `SelectMessage.selectors` is a tuple, not a list.
- `FunctionAnnotation` is flattened into `Expression.function` and `Expression.options`.

As far as I can tell, no external users should be impacted by these changes.


mathjazz (approved):
I didn't run the code, but nothing in the changes and tests stands our. r+

PR #41, opened 2025-01-22
Add a compact JSON representation for messages

eemeli:
~This builds on #40, and may need a rebase after that's merged. Actual diff here is [`add-mf2...msg-data-json`](https://github.com/mozilla/moz-l10n/compare/add-mf2...msg-data-json).~ _Done._

Adds converters to and from a compact JSON-serializable representation of a `Message`.

The format of the output is defined by the `/moz/l10n/message/schema.json` JSON Schema.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,37 @@ and/or contents.
 
 Returns a `Format` enum value, or `None` if the input is not recognized.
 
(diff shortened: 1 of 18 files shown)


mathjazz (approved):
This is a bit abstract, but looks good.

PR #43, opened 2025-01-23
Remove setup.cfg (only used for obsolete flake8)

flodolo:


flodolo:
I'm assuming this is not used anymore. Feel free to close if that's not the case.

eemeli (approved):

PR #45, opened 2025-01-27
Force encoding when serializing resources

serge-sans-paille:
The default encoding is platform dependent and this can cause issue on windows, see https://bugzilla.mozilla.org/show_bug.cgi?id=1942257


eemeli (approved):

eemeli:
@serge-sans-paille The current `main` has some breaking changes, but I'll release an off-branch 0.5.7 with this quite soon.

PR #50, opened 2025-02-18
Move Python implementation to `python/`

eemeli:
As discussed in #46, to match the JS implementation having its own `js/` directory.

Dev dependencies and lockfiles of both implementations are now in the root, to make setup & development easier.

A new root README is added, with dev setup instructions.

One of the Python tests needed a fix to make `pytest` work when run from either the repo root or the `python/` directory.


eemeli:
Took a few iterations to get the CI commands right. Looks like `mypy` will only succeed when run from the root, and not from `python/`.

flodolo (approved):

PR #51, opened 2025-02-19
Allow `//` line comments in messages.json

### LINKED ISSUE #47: webext JSON comments
Hi @eemeli I wanted to check if moz-l10n would address some long outstanding issues we have with some formats in Weblate, and it turns out that comments in webext JSON are not handled here as well.

The issue is in more detaile described at https://github.com/WeblateOrg/weblate/issues/2043, but to make it short:

> It is a [JSON](https://developer.mozilla.org/en-US/docs/Glossary/JSON)-formatted file, with one exception: it is allowed to contain "//"-style comments.

This is something that `json.loads` cannot handle:

https://github.com/mozilla/moz-l10n/blob/9f8035997af7926e60f4ad694483ce67e06290a4/moz/l10n/formats/webext/parse.py#L47

eemeli:
Closes #47, ping @nijel

As discussed in the issue, this discards `//` line comments when parsing a file with them, and does not include them in the serialization.

The regular expression has been adapted from the one [used by Firefox](https://searchfox.org/mozilla-central/rev/0f754bb4607893ad58df525601334a18f9650d57/toolkit/components/extensions/Extension.sys.mjs#398).

At least for now, comments in JSON are not supported for the `plain_json` format.

### DIFF
--- a/python/moz/l10n/util/loads.py
+++ b/python/moz/l10n/util/loads.py
@@ -1,0 +1,32 @@
+# Copyright Mozilla Foundation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import annotations
+
+from json import loads
+from re import MULTILINE, compile
+from typing import Any
+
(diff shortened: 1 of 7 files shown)


nijel:
Pretty much matches what I [ended up doing in translate-toolkit](https://github.com/translate/translate/pull/5486) :slightly_smiling_face:

python/moz/l10n/util/loads.py
flodolo, line 1: Is only my brain confused by the fact that you now have a `loads` module, which loads a `json.loads` as `loads`? Should there be some differentiation, at least on this file name?

flodolo (approved):

python/moz/l10n/util/loads.py
eemeli: Python at least is pretty clear about what this means, as the name of the current module isn't referenced anywhere within the module itself. I picked a different name than the full function name so that code completion doesn't get confused, as it sometimes does with modules like that.

PR #52, opened 2025-02-19
Add paths.all_locales

### LINKED ISSUE #49: It should be possible to access the superset of locales from all included config files
`compare-locales` had the concept of [all_locales](https://github.com/mozilla/compare-locales/blob/41e148f6ed6416f65f0e4592169b4745853f52cf/compare_locales/paths/project.py#L139)

That's something we use, for example, in Android
https://github.com/mozilla-l10n/android-l10n/blob/main/firefox.toml
https://github.com/mozilla-l10n/android-l10n/blob/main/focus.toml

eemeli:
Fixes #49 by adding a property `all_locales` with a similar signature and behaviour as in compare-locales. Note that the value returned here is a set, rather than a sorted list.

### DIFF
--- a/python/moz/l10n/paths/config.py
+++ b/python/moz/l10n/paths/config.py
@@ -208,6 +208,24 @@ class L10nConfigPaths:
             incl.locales = locales
 
     @property
+    def all_locales(self) -> set[str]:
+        """
+        All locales for the config,
+        including locales set in its imports and paths.
+        """
+        return set(self._all_locales)
+
+    @property
+    def _all_locales(self) -> Iterator[str]:
+        if self._locales is not None:
+            yield from self._locales
+        for _, locales in self._path_data.values():
+            if locales is not None:
+                yield from locales
+        for incl in self._includes:
+            yield from incl._all_locales
+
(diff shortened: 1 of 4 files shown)


python/moz/l10n/paths/config.py
```diff
@@ -207,6 +207,24 @@ def locales(self, locales: list[str] | None) -> None:
         for incl in self._includes:
             incl.locales = locales
 
+    @property
+    def all_locales(self) -> set[str]:
+        """
+        All locales for the config,
+        including locales set in its imports and paths.
```
flodolo, line 214: Why `and paths`?

flodolo (approved):
Tested and it works as expected 👍🏼

python/moz/l10n/paths/config.py
eemeli: Because that's what compare-locales also does:
https://github.com/mozilla/compare-locales/blob/41e148f6ed6416f65f0e4592169b4745853f52cf/compare_locales/paths/project.py#L146-L148

PR #54, opened 2025-02-19
Always use Message as Entry.value

eemeli:
Resources are easier to work with if their structure is fixed, and `Entry.value` is always a `Message`. Currently, this optionality is only used by `fluent_parse(..., as_ftl_patterns=True)`, and it adds a bit of complexity everywhere.

Most of the code changes here are in the serialisers, which no longer need to support Resources with string values.


mathjazz (approved):
Nothing stands out. r+

PR #55, opened 2025-02-25
Add _ prefix to all internal modules

eemeli:
The library consists of a number of modules, many of which are not intended for external users. Adding a `_` prefix for the names of modules that are not intended to be a part of the public API makes it clearer what that public API _is_, and what may of may not change without being considered a breaking change.

In JS, this is done much more easily with an explicit package.json exports field: https://github.com/mozilla/moz-l10n/blob/53ad5a6d0211b5c68a6c01ae0ffe3eb59a9ac23a/js/package.json#L23-L26


python/moz/l10n/bin/build.py
```diff
@@ -24,7 +24,7 @@
 
 from moz.l10n.formats import Format
 from moz.l10n.model import Comment, Entry, Resource, Section
-from moz.l10n.paths.config import L10nConfigPaths
+from moz.l10n.paths import L10nConfigPaths
```
flodolo, line 27: I get the other changes, but how this (and the change in `fix.py`) work, since the class didn't move?

eemeli: They're re-exported from the __init__ module.

eemeli:
Never mind, this messes up Pylance in projects using moz.l10n.

PR #58, opened 2025-02-28
Various resource format fixes

eemeli:
Identified during Pontoon dev work:
- Gettext plural messages are now consistently parsed & serialised with the same numerical identifiers.
- A gettext `msgctx` is considered secondary to the `msgid`, rather than a section header.
- When serialising, consider entries with `str` values to have already been appropriately escaped.
- Extraneous string contents in Android strings.xml files should not cause an error.
- In XLIFF, be consistent regarding empty `<node>` and `<target>` elements.

### DIFF
--- a/python/moz/l10n/formats/po/serialize.py
+++ b/python/moz/l10n/formats/po/serialize.py
@@ -15,6 +15,7 @@
 from __future__ import annotations
 
 from collections.abc import Iterator
+from re import match
 
 from polib import POEntry, POFile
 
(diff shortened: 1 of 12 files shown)


python/moz/l10n/formats/po/serialize.py
```diff
@@ -24,35 +25,41 @@
 def po_serialize(
     resource: Resource[str] | Resource[Message],
     trim_comments: bool = False,
-    wrapwidth: int = 78,
+    wrapwidth: int = 200,
```
mathjazz, line 28: I assume we use 200 here because we use it at Mozilla broadly?

Non-blocking: It would be nice if this were configurable.

mathjazz (approved):

PR #59, opened 2025-03-03
Add single-message/pattern parsers

eemeli:
Single-message Python parsers are required by Pontoon's data model migration. A general-purpose `parse_message()` is added for all formats except for Fluent, for which a seprate `fluent_parse_message()` is needed because of the source representing a set of entries, rather than a single pattern.

While at this, support for Xcode message patterns is improved, with a less hacky detector that does not depend on the file name but the `tool-id` that's set within its contents.

As a simplification, the Fluent parser & serialiser will no longer support mixed resources, i.e. ones where a resource contains FTL AST messages.

### DIFF
--- a/python/README.md
+++ b/python/README.md
@@ -116,6 +116,34 @@ def message_to_json(msg: Message) -> list[Any] | dict[str, Any]: ...
 Converters to and from a JSON-serializable representation of a `Message`.
 The format of the output is defined by the [`schema.json`](./moz/l10n/message/schema.json) JSON Schema.
 
+### moz.l10n.message.parse_message
+
+```python
+from moz.l10n.message import parse_message
+def parse_message(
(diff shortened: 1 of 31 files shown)


mathjazz:
> Single-message Python parsers are required by Pontoon's data model migration. A general-purpose `parse_message()` is added for all formats except for Fluent, for which a seprate `fluent_parse_message()` is needed because of the source representing a set of entries, rather than a single pattern.

To avoid any confusion - you meant `fluent_parse_messages()`, right?

mathjazz (approved):
Looks good. 

(I'm curious how this will be used in Pontoon given the specifics of Fluent.)

eemeli:
Yes, messages plural for Fluent. The format will indeed need special handling in Pontoon.

PR #60, opened 2025-03-04
paths: format_target_path should not re-apply base directory to path

eemeli:
Fixes https://github.com/mozilla-l10n/android-l10n/pull/888#issuecomment-2695150112


flodolo (approved):
Thanks, this works with the code you suggested in android-l10n (will need a new release though)

PR #62, opened 2025-03-10
Ignore linepos when comparing entries and sections

### LINKED ISSUE #36: add_entries: Line positions should be ignored in comparisons
Since including line positions in 0.4.0 and later, [this comparison](https://github.com/mozilla/moz-l10n/blob/4fcaf5c01dec938d7986a214c5bb4551745123ef/moz/l10n/resource/add_entries.py#L66) in `add_entries()` is considering entries to differ even if the difference is only in their exact position within the source file.

flodolo:
Fixes #36 

Not sure if I'm oversimplifying things (do we care about `linepos` at call when doing comparisons?).

### DIFF
--- a/python/moz/l10n/model.py
+++ b/python/moz/l10n/model.py
@@ -304,7 +304,7 @@ class Section(Generic[V]):
     Metadata attached to this section.
     """
 
-    linepos: LinePos | None = None
+    linepos: LinePos | None = field(default=None, compare=False)
     """
     The parsed position of the section,
     available for some formats.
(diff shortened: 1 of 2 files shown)


python/moz/l10n/model.py
```diff
@@ -248,7 +248,7 @@ class Entry(Generic[V]):
     Metadata attached to this entry.
     """
 
-    linepos: LinePos | None = None
+    linepos: LinePos | None = field(default=None, compare=False)
```
eemeli, line 251: This same change ought to be also applied to line 307 for Section.

eemeli (changes requested):

PR #65, opened 2025-04-02
Add single-message serializers

eemeli:
I'd rather hoped to avoid needing to provide single-message serialisers, but it turns out that they're effectively required for the Pontoon data model migration -- without them, we'd need to parse each resource using both the old & new tools, and to merge their representations rather clumsily.

With these serialisers, we can instead replace the current Pontoon parsers with the moz.l10n ones, and produce the current resource representations using moz.l10n, allowing for a smaller next step in Pontoon.

Thankfully, these serialisers were even easier to extract from pre-existing code than the parsers of #59. Fluent is again left out, as it'll need special treatment, using the already provided `fluent_astify_message()`.

### DIFF
--- a/python/README.md
+++ b/python/README.md
@@ -109,8 +109,9 @@ Returns a `Format` enum value, or `None` if the input is not recognized.
 ```python
 from moz.l10n.message import from_json, to_json
 
-def message_from_json(json: list[Any] | dict[str, Any]) -> Message: ...
-def message_to_json(msg: Message) -> list[Any] | dict[str, Any]: ...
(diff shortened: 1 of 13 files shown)


flodolo:
Given these changes, would it be possible to access the string representation of the message as one of the message's properties (like `raw_val` in compare-locales)?

eemeli:
No, because the serialization depends on the format, which is stored in the resource, and a message does not contain a reference to its resource. It would also be quite wasteful to pre-compute, as it's not required in most cases.

mathjazz (approved):

PR #66, opened 2025-04-02
Miscellaneous fixes from Pontoon work

eemeli:
See individual commits.

### DIFF
--- a/python/moz/l10n/formats/__init__.py
+++ b/python/moz/l10n/formats/__init__.py
@@ -80,12 +80,14 @@ class UnsupportedFormat(Exception):
     pass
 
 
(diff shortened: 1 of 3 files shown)


eemeli:
The metadata structure is generic but minimal (a list of key-value pairs with no uniqueness contraints), so that it can represent all sorts of different information that formats like to store about entries, sections, and resources. With `get_meta(key)`, it becomes easier to use it as a key-value store, which is relatively common.

For example`entry.get_meta('fuzzy') == 'true'` can now be used to determine if a .po entry is fuzzy or not.

mathjazz:
Thanks for clarifying. I was reading the docstring too fast. I missed the "with a matching `key`" in "First metadata entry with a matching `key`".

PR #68, opened 2025-04-07
paths: Add locale argument to target()

eemeli:
Fixes (but does not close) mozilla/pontoon#3572

Allows for multiple paths configs with the same reference path but different locales.

The lookups in `L10nConfigPaths` `target()` and `find_reference()` now need to iterate through a list, rather than doing a `key in dict` lookup.

A `locale` argument is added to `target()` to allow for the right path to be found, and for formatting to be applied. The idea here is that if you're looking for a specific locale's target path, you'll want the formatted path.

While technically this isn't a breaking change, the calling code in Pontoon will need a review to make sure that isn't making any similar assumptions.

### DIFF
--- a/python/moz/l10n/paths/config.py
+++ b/python/moz/l10n/paths/config.py
@@ -118,10 +118,11 @@ class L10nConfigPaths:
         which are also the indexed groups captured in `target`.
         """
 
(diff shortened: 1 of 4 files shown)


mathjazz (approved):
Looks good.

We should check how this change affect performance in Pontoon.

PR #78, opened 2025-05-22
Use hexadecimal rather than decimal `\u` values in Android string resources

eemeli:
I was previously misled by [this section](https://developer.android.com/guide/topics/resources/string-resource#escaping_quotes) of the string resource docs:
> Whitespace collapsing and Android escaping happens after your resource file gets parsed as XML. This means that `<string> &#32; &#8200; &#8195;</string>` (space, punctuation space, Unicode Em space) all collapse to a single space (`" "`), because they are all Unicode spaces after the file is parsed as an XML. To preserve those spaces as they are, you can either quote them (`<string>" &#32; &#8200; &#8195;"</string>`) or use Android escaping (`<string> \u0032 \u8200 \u8195</string>`). 

That last example there should read instead: `<string> \u0020 \u2008 \u2003</string>`.

### DIFF
--- a/python/moz/l10n/formats/android/parse.py
+++ b/python/moz/l10n/formats/android/parse.py
@@ -438,7 +438,7 @@ def parse_quotes(
 
 
 inline_re = compile(
-    r"\\u([0-9]{4})|"
+    r"\\u([0-9a-fA-F]{4})|"
     r"\\(.)|"
     r"(<[^%>]+>)|"
     r"(%(?:[1-9]\$)?[-#+ 0,(]?[0-9.]*([a-su-zA-SU-Z%]|[tT][a-zA-Z]))"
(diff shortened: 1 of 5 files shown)


flodolo:
What is the expected behavior on trailing and starting white spaces? One of the strings where we ended up with these `\u0032` is [this one](https://searchfox.org/mozilla-central/source/mobile/android/fenix/app/src/main/res/values/strings.xml#64-66).

en-US

```
    <string name="private_browsing_common_myths">
       Common myths about private browsing
    </string>
```

en-GB (one of many examples)

```
  <string name="private_browsing_common_myths">\u0032Common myths about private browsing\u0032</string>
```

eemeli:
According to the paragraph I quoted above, all whitespace gets collapsed to a single space, but is otherwise retained. But it's of course possible that that's wrong as well, and now I need to investigate that too. I'll file a separate PR if that's necessary.

The intent with the escaped leading & trailing space is to indicate that they are indeed there.

PR #102, opened 2025-09-17
android: Normalize whitespace before & after some tags as a newline

eemeli:
As whitespace around HTML block tags in Android will eventually get swallowed up, the space character that we use has no impact on the formatted results. Hence, let's ensure that block tags always start on a new line, so that the messages are easier to read and translate.


mathjazz (commented):
Please fix linting.

PR #135, opened 2026-05-06
Add message normalization

eemeli:
Adds JS & Python functions for message normalization, i.e. dropping unused declarations, joining adjacent literal elements, and dropping empty literal elements.

The intent is to use this in Pontoon to ensure that translations that _appear_ to be the same are also considered equal.

### DIFF
--- a/js/README.md
+++ b/js/README.md
@@ -25,6 +25,22 @@ and the parser for `mf2` depends on the `messageformat` package.
 
 ## API
 
+### normalizeMessage()
+
+```js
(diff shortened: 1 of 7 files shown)


mathjazz (approved):
JS `normalizeMessage()` clones and does not modify input, but Python `.normalize()` mutates and returns self. That may be intentional, but should be called out in Python docs/tests to avoid surprises.

Other than that, this is high-quality work that will make equality/comparison logic much more reliable. Great addition to the library!

Issue #37, opened 2024-12-19
Improve L10nDiscoverPaths performance


eemeli:
Path discovery seems to take a bit too long when run on the [firefox-l10n-source](https://github.com/mozilla-l10n/firefox-l10n-source) and [firefox-l10n](https://github.com/mozilla-l10n/firefox-l10n), which contain about 44k localization files.

It should be faster.

eemeli:
This is the bit of code that's way too slow: https://github.com/mozilla/moz-l10n/blob/3f6e98e5bbf03cbdd76586da36932464a1af813c/moz/l10n/paths/discover.py#L163-L165

For Firefox, `len(base_dirs) == 730` and `len(l10n_dirs) == 11046`.

ewerybody:
I'd like to have a look into this! :)

ewerybody:
The stats are in! I put in some prints and originally we had:
```
walking took 1.473s
cleaning ref_root took 0.000s
filter ref dirs took 0.070s
pick base dirs1 took 0.005s
pick base dirs2 took 21.803s
rest took 0.018s
all took 23.373s
```
With more than 4 milion calls to `dir_contains`!!! 
So it was looking to map each of the 737 `base_dirs` to each of the 11005 `l10n_dirs`.
737 * 11005 are actually 8110685! so why "only" 4109525? The `any` in [this line 164^](https://github.com/mozilla/moz-l10n/blob/3f6e98e5bbf03cbdd76586da36932464a1af813c/moz/l10n/paths/discover.py#L163-L165) was early exiting half the time.

Now I flipped it a bit upside down, going over each of the `l10n_dirs` and look up a couple parent directories to see if it maps to one of the `base_dirs`. Way less and less heavy lookups ops. We get:
```
walking took 1.528s
cleaning ref_root took 0.000s
filter ref dirs took 0.075s
pick base dirs1 took 0.004s
pick base dirs2 took 0.068s
rest took 0.021s
all took 1.700s
```
I serialized the results of `L10nDiscoverPaths` on [firefox-l10n-source](https://github.com/mozilla-l10n/firefox-l10n-source) and [firefox-l10n](https://github.com/mozilla-l10n/firefox-l10n), before and after and got the exact same!👍

PR #137, opened 2026-05-15
Improves `L10nDiscoverPaths` performance

ewerybody:
Addressing #37

- changed heavy `base_dirs` loop for checks against hashed sets.
- flipped `.pot`-dirs lookup to after `l10n_extensions` check
- added `test_split_repos` mocking `firefox-l10n-source/firefox-l10n` situation.
- changed `build_file_tree` in `utils` to use `makedirs` to be able to create dirs in yet inexistent paths.

### DIFF
--- a/python/moz/l10n/paths/discover.py
+++ b/python/moz/l10n/paths/discover.py
@@ -18,7 +18,7 @@ import re
 from collections.abc import Iterable, Iterator
 from itertools import chain
 from os import sep, walk
-from os.path import commonpath, isdir, join, normpath, relpath, splitext
+from os.path import commonpath, dirname, isdir, join, normpath, relpath, splitext
 
 from moz.l10n.formats import l10n_extensions
 from moz.l10n.util import walk_files
@@ -130,13 +130,17 @@ class L10nDiscoverPaths:
                 base_dirs.append((dirpath, locale_dirs))
             dirnames[:] = (dn for dn in dirnames if not dn.startswith("."))
 
-            if any(not fn.startswith(".") and fn.endswith(".pot") for fn in filenames):
-                pot_dirs.append(dirpath)
-            if any(
+            # .pot is in `l10n_extensions`!
+            # We don't need to check for these if there isn't any at all.
+            if not any(
                 not fn.startswith(".") and splitext(fn)[1] in l10n_extensions
                 for fn in filenames
             ):
-                l10n_dirs.append(dirpath)
+                continue
+
+            l10n_dirs.append(dirpath)
+            if any(not fn.startswith(".") and fn.endswith(".pot") for fn in filenames):
+                pot_dirs.append(dirpath)
 
         if ref_root:
             for dir in list(ref_dirs):
@@ -160,8 +164,19 @@ class L10nDiscoverPaths:
         # with the most locale subdirectories,
         # with a preference for directories with localizable contents.
         base_dirs = [bd for bd in base_dirs if not dir_contains(self._ref_root, bd[0])]
+        base_dir_roots = {bd[0] for bd in base_dirs}
+        confirmed_base_dirs: set[str] = set()
+        for ld in l10n_dirs:
+            parent = ld
+            while True:
+                parent = dirname(parent)
+                if parent in base_dir_roots:
+                    confirmed_base_dirs.add(parent)
+                    break
+                if parent == root or parent == dirname(parent):
+                    break
         base_dirs = [
-            bd for bd in base_dirs if any(dir_contains(bd[0], ld) for ld in l10n_dirs)
+            bd for bd in base_dirs if bd[0] in confirmed_base_dirs
         ] or base_dirs
 
         locale_dirs_: list[str] | None
--- a/python/moz/l10n/paths/discover.py
+++ b/python/moz/l10n/paths/discover.py
@@ -172,7 +172,6 @@ class L10nDiscoverPaths:
                 parent = dirname(parent)
                 if parent in base_dir_roots:
                     confirmed_base_dirs.add(parent)
-                    break
                 if parent == root or parent == dirname(parent):
                     break
         base_dirs = [
--- a/python/moz/l10n/paths/discover.py
+++ b/python/moz/l10n/paths/discover.py
@@ -110,7 +110,6 @@ class L10nDiscoverPaths:
 
         # dir -> score
         ref_dirs: dict[str, int] = {ref_root: len(source_locale)} if ref_root else {}
-        # base_dirs: list[tuple[str, list[str]]] = []  # [(root, [locale_dir])]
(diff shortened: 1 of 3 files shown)


flodolo:
Let's wait for @eemeli, but I might have stumbled upon a bug — or at least different behavior — while asking Claude to experiment with benchmark tests.

```python
    def test_split_repos_locale_id_intermediate_names(self):
        source_tree: Tree = {"dom": {"aa": {"strings.ftl": ""}}}
        l10n_tree: Tree = {
            "de": {"dom": {"aa": {"strings.ftl": ""}}},
            "fr": {"dom": {"aa": {"strings.ftl": ""}}},
        }
        with TemporaryDirectory() as tmp:
            source_root = join(tmp, "src")
            l10n_root = join(tmp, "l10n")
            build_file_tree(source_root, source_tree)
            build_file_tree(l10n_root, l10n_tree)

            paths = L10nDiscoverPaths(l10n_root, ref_root=source_root)
            assert paths.ref_root == source_root
            assert paths.base == l10n_root
            assert paths.locales is not None
            assert set(paths.locales) == {"de", "fr"}
```

This doesn't break with the real firefox-l10n repository because normally there is a file in the root of the locale folder.

ewerybody:
Oh OK! Yeah sure! I'm curious what he'll say :)
Thanks for looking into it!

eemeli:
Hey, I'm travelling this week so probably won't have time to properly look at this until later.

ewerybody:
@flodolo can you check back again? I think I found and fixed that issue!
I guess I was a little too eager breaking the loop early on **any** find. So when we simply let it go deeper that resolves it! :)

flodolo:
Yes, this seems to work. Performance is worse compared to the original fix, but it still seems better than the currenct code (might be worth checking the numbers).

ewerybody:
Yeah removing that eager `break` might yield way more loops.
But I still get 1.690s vs the original 23.373s on the firefox-l10n-source / firefox-l10n repos. 🤞

python/moz/l10n/paths/discover.py
```diff
@@ -130,13 +130,17 @@ def __init__(
                 base_dirs.append((dirpath, locale_dirs))
             dirnames[:] = (dn for dn in dirnames if not dn.startswith("."))
 
-            if any(not fn.startswith(".") and fn.endswith(".pot") for fn in filenames):
-                pot_dirs.append(dirpath)
-            if any(
+            # .pot is in `l10n_extensions`!
+            # We don't need to check for these if there isn't any at all.
+            if not any(
```
eemeli: The note about `.pot` is not necessary, and it's cleaner if we don't invert the condition and `continue`, but apply the action if the condition is fulfilled.
```suggestion
            if any(
```

python/moz/l10n/paths/discover.py
```diff
@@ -130,13 +130,17 @@ def __init__(
...
                 not fn.startswith(".") and splitext(fn)[1] in l10n_extensions
                 for fn in filenames
             ):
-                l10n_dirs.append(dirpath)
+                continue
+
+            l10n_dirs.append(dirpath)
+            if any(not fn.startswith(".") and fn.endswith(".pot") for fn in filenames):
+                pot_dirs.append(dirpath)
```
eemeli:
```suggestion
                l10n_dirs.append(dirpath)
                if any(not fn.startswith(".") and fn.endswith(".pot") for fn in filenames):
                    pot_dirs.append(dirpath)
```

python/moz/l10n/paths/discover.py
```diff
@@ -157,11 +161,22 @@ def __init__(
             raise MissingSourceDirectoryError
 
         # Pick the localization base dir not in the reference directory
-        # with the most locale subdirectories,
-        # with a preference for directories with localizable contents.
```
eemeli, line 161: The preference for directories with localizable contents is still present, and the note about it should not be dropped as it's not necessarily obvious.

python/moz/l10n/paths/discover.py
```diff
@@ -157,11 +161,22 @@ def __init__(
...
+        # with the most locale subdirectories.
+        # Walk up from each l10n_dir to confirm which base dirs actually contain
+        # localizable files and avoid false positives on locale-id-like dir names.
         base_dirs = [bd for bd in base_dirs if not dir_contains(self._ref_root, bd[0])]
+        base_dir_roots = {bd[0] for bd in base_dirs}
+        confirmed_base_dirs: set[str] = set()
```
eemeli: It would probably be more straightforward to turn the `base_dirs` list-of-tuples into a dict, and then to use the loop to build a filtered copy of it that can then be used below if it's not empty.

I think I may have used a list-of-tuples instead of a dict in the first place only to ensure that the insertion order is retained, as that matters for the `max()` call below.

Regarding names, It's not clear what is "confirmed" here.

eemeli (changes requested):
Overall looks good; not surprised that there's an N * M problem here.

python/moz/l10n/paths/discover.py
```diff
@@ -156,17 +159,37 @@ def __init__(
...
+        # subdirectories, with preference for directories with localizable contents.
+        # Walk up from each l10n_dir to confirm which base dirs actually contain
+        # localizable files and avoid false positives on locale-id-like dir names.
+        # base_dirs = [bd for bd in base_dirs if not dir_contains(self._ref_root, bd[0])]
+        base_dirs = {
+            bd: lds
+            for bd, lds in base_dirs.items()
+            if not dir_contains(self._ref_root, bd)
+        }
+        found_parent_dirs: set[str] = set()
```
eemeli: It's still not clear what "found" means here. Maybe something `_with_contents`?

Also, as the only reason we're not directly constructing the dict inside the loop is to keep the items in the order that gets us the shortest path first, either:
1. that should be commented upon as it's not obvious (and I should have done so earlier), or
2. we should let the order change here by building the dict directly, and update the `max` key lambda to have a secondary dependence on the number of path parts, as we want to select the shortest path.

python/moz/l10n/paths/discover.py
```diff
@@ -156,17 +159,37 @@ def __init__(
...
+                if parent == root or parent == dirname(parent):
+                    break
+        base_dirs = {
+            bd: lds for bd, lds in base_dirs.items() if bd in found_parent_dirs
+        } or base_dirs
+        base_dirs = {
+            base_dir: locale_dirs
+            for base_dir, locale_dirs in base_dirs.items()
+            if not dir_contains(self._ref_root, base_dir)
+        }
```
eemeli: This is mistakenly included a second time, right?

eemeli (changes requested):
See inline, and the now-commented lines should be dropped.

eemeli (approved):
🎉

PR #146, opened 2026-05-27
Improve Android placeholder handling

eemeli:
Applies a number of Android's [java.util.Formatter](https://developer.android.com/reference/java/util/Formatter) practices to provide a better match with runtime behaviour.

For [`%<` references](https://developer.android.com/reference/java/util/Formatter#argument-index) to the preceding argument's index, the referenced argument index is used internally and in any subsequent serialization. This is to enable moving the placeholders around so their order changes during translation.

Now all `%` characters are presumed to start a placeholder, which matches what's done with Android strings that are formatted with placeholders. But this does reveal a slight problem: An Android `strings.xml` file does not contain sufficient information to determine whether a string is to be formatted with placeholders or not, and if it's not, then if it contains a literal non-terminal `%`, that can get mis-handled. This should not be a problem with any of our current messages, but we ought to handle this a bit better, possibly with heuristics to detect the string format (which could theoretically also be ICU MessageFormat, or anything else) and/or a requirement to specify the message format as a `<string>` attribute.

### DIFF
--- a/js/src/android-parse.ts
+++ b/js/src/android-parse.ts
@@ -166,9 +166,9 @@ function* parseQuotes(
 }
 
 const _inline =
-  // 1:esc-unicode     2:esc-char       4:printf
-  //                         3:esc-elem                                  5:printf-conversion              6:xml-entity
-  /\\u([0-9A-Fa-f]{4})|\\(.)|(<[^%>]+>)|(%(?:[1-9]\$|<)?[-#+ 0,(]?[0-9.]*([a-su-zA-SU-Z%@]|[tT][a-zA-Z]))|(_entity_\d+_)/g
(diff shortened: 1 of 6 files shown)


flodolo (commented):
The Python part makes sense to me (had no idea about `%n` and `%t`).

flodolo (approved):

eemeli:
Note that unlike `\t`, `%t` is not a valid escape for a tab character; it's a [prefix for date and time conversion](https://developer.android.com/reference/java/util/Formatter.html#:~:text=Prefix%20for%20date%20and%20time%20conversion). That's unlike `\n` and `%n`, which are escapes for newline characters.

PR #147, opened 2026-05-28
Add translation coverage reporting to l10n-build and l10n-build-file

bakulf:
## Summary

Adds optional per-locale translation coverage reporting to the build tools.

### l10n-build
- New `--coverage` flag. When set, writes a `coverage.json` per locale under
  `--target`, mapping each file to its message coverage:
  ```json
  {
    "file.ftl": { "total": 4, "missing": ["msg-c", "msg-d"] }
  }
- Message ids are kept structural: a single-part id is emitted as a string,
while multi-part ids (are emitted as a list of strings.

### l10n-build-file

- New `--coverage PATH` option, mirroring l10n-build's capability for the
single-file builds.

### DIFF
--- a/python/moz/l10n/bin/build.py
+++ b/python/moz/l10n/bin/build.py
@@ -14,6 +14,7 @@
 
 from __future__ import annotations
 
+import json
 import logging
 from argparse import ArgumentParser, RawDescriptionHelpFormatter
 from collections import defaultdict
@@ -61,6 +62,11 @@ def cli() -> None:
     parser.add_argument(
         "--locales", metavar="LOCALE", nargs="+", required=True, help="target locales"
     )
+    parser.add_argument(
+        "--coverage",
+        action="store_true",
+        help="write a coverage.json file per locale with the translation ratio",
+    )
     args = parser.parse_args()
 
     log_level = (
@@ -76,9 +82,17 @@ def cli() -> None:
     l10n_base: str = args.base
     l10n_target: str = args.target
     locales: set[str] = set(args.locales)
+    write_coverage: bool = args.coverage
 
     # locale -> [ftl_missing, src_fallback]
     msg_data: dict[str, list[int]] = defaultdict(lambda: [0, 0])
+
+    # locale -> {file_path -> {"total": int, "missing": [id, ...]}}.
+    # Pre-initialized so every requested locale gets a coverage.json,
+    # even if no source files were parseable.
+    coverage_data: dict[str, dict[str, dict[str, int | list[str | list[str]]]]] = {
+        locale: {} for locale in locales
+    }
 
     paths = L10nConfigPaths(cfg_path)
     paths.base = l10n_base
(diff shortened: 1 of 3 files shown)


eemeli (changes requested):
I have not yet looked at the changes in any detail, as at least two issues need to be resolved first:
1. This is missing a description of why we're applying these changes. We've been good so far to have reasons why we're adding any new capabilities to this library, and that should continue.
2. As I've mentioned on a number of times previously, the coverage output we want to emit here should be a simple list of files which do not have sufficient coverage. The determination of what constitutes "sufficient coverage" needs to be made here, rather than at any later time.

bakulf:
After discussing this PR on slack, the new solution is to generate a `coverage.json` file with the fluent file relative path as an index for objects containing: total number of strings, an array of missing IDs.

python/moz/l10n/bin/build.py
```diff
@@ -164,13 +188,16 @@ def get_entry(
                         }
                         return l10n_entry
                     msg_delta -= 1
+                    missing_ids.append(".".join(id))
```
eemeli: The id should not be collapsed into a string like this; doing so makes its processing format-specific. For example, `.` is a rather common character in .properties message keys.

If there is a desire to simplify, then an id with a single part could be represented as a string, rather than an array with a single string value.

eemeli (changes requested):
Looks pretty good, though a few things need to be addressed:

- [x] The PR title and description need to be updated.
- [x] The coverage should only get written if a `--coverage` CLI argument is used.
- [x] There's also `build_file.py`, which gets [called by fastermake](https://searchfox.org/firefox-main/rev/e28b34ab33dbf49364999070168cbb7e11e8e5bd/python/mozbuild/mozbuild/backend/fastermake.py#226) and should probably match `build.py` in capabilities?
- [x] Message `id` tuples should not be always joined into a single string; doing so makes their processing format-specific.

python/moz/l10n/bin/build_file.py
```diff
@@ -75,8 +84,27 @@ def cli() -> None:
...
+                args.source, source_res, args.l10n, args.target
+            )
+
+        if args.coverage:
+            coverage_dir = dirname(args.coverage)
+            if coverage_dir:
+                makedirs(coverage_dir, exist_ok=True)
+            with open(args.coverage, "w", encoding="utf-8") as file:
+                json.dump(
+                    {"total": total_count, "missing": missing_ids},
```
eemeli: We need to use the same file format as for the l10n-build coverage.json, i.e. one file per locale for all the resources. So when processing one file at a time with fastermake, we need to read any pre-existing coverage.json (or create it as empty if missing) and overwrite (or add) the entry of the current resource.

eemeli (changes requested):

python/moz/l10n/bin/build_file.py
bakulf: I used the 'target' as key of the object. Let me know if you prefer the baseline or having another param to define the object key.

python/moz/l10n/bin/build_file.py
```diff
@@ -75,8 +84,33 @@ def cli() -> None:
...
+
+        if args.coverage:
+            # Match l10n-build: one coverage.json per locale, keyed by the
+            # target path. Read any pre-existing file and add/overwrite this
+            # resource's entry.
+            coverage_data: dict[str, dict[str, object]] = {}
+            if exists(args.coverage):
+                with open(args.coverage, encoding="utf-8") as file:
+                    coverage_data = json.load(file)
+            coverage_data[args.target] = {
```
eemeli: This does not seem right. For l10n-build you have
https://github.com/mozilla/moz-l10n/blob/8a5b1a2e70c978402f24ea3b449f2bc16c458b38/python/moz/l10n/bin/build.py#L122-L126
and `args.target` won't always match that, esp. as relative/absolute paths are not normalised.

The tests probably need to include a case where a coverage.json file is first created with l10n-build, and then updated with l10n-build-file.

We'll probably need to revisit exactly what path we're using here as the runtime consumer components are developed further, as they'll need to be able to map from e.g. a Firefox Fluent request path like `browser/browser.ftl` to a coverage data path. But that's a later concern; initially let's just try and keep the l10n-build and l10n-build-file behaviours matching each other.

eemeli (changes requested):

python/moz/l10n/bin/build_file.py
bakulf: The latest version of this patch introduces a `--coverage-name` parameters to define the key.

eemeli (approved):
This looks good! Let's go with this, and (probably) iterate as we get more experience with its usage.

PR #148, opened 2026-06-02
Fix parsing & serialization of empty patterns

eemeli:
Ensure that empty patterns are parsed as `[]` rather than `['']`, and that they serialize correctly.


mathjazz (approved):
The `[]` representation for empty patterns is consistent across JS and Python, and the new round-trip tests cover the affected formats. No blockers from my side.

PR #149, opened 2026-06-04
Add compatibility with lxml 6.*

flodolo:
This follows the previous pattern, leaving compatibility also with 6.0.x (I assume we don't want to drop that).


python/pyproject.toml
```diff
@@ -39,7 +39,7 @@ dependencies = [
 ]
 
 [project.optional-dependencies]
-xml = ["lxml >= 6.0.0, < 6.1.0"]
+xml = ["lxml >= 6.0.0, < 6.2.0"]
```
eemeli: This certainly works, but maybe we ought to be a bit more lenient with the range? We don't do very complicated things with `lxml`, and it's unlikely to break semver.

```suggestion
xml = ["lxml >= 6.0.0, < 7.0.0"]
```

eemeli (approved):

uv.lock
```diff
@@ -318,7 +312,7 @@ requires-dist = [
     { name = "fluent-syntax", specifier = "~=0.19.0" },
     { name = "gitignorant", specifier = "~=0.3.1" },
     { name = "iniparse", specifier = "~=0.5" },
-    { name = "lxml", marker = "extra == 'xml'", specifier = ">=6.0.0,<6.1.0" },
+    { name = "lxml", marker = "extra == 'xml'", specifier = ">=6.0.0,<6.2.0" },
```
eemeli: I think this also needs an update.

flodolo: On the phone, so I won't be able to look before next week. Feel free to take over and fix if you want to get it out of the way.

eemeli:
```suggestion
    { name = "lxml", marker = "extra == 'xml'", specifier = ">=6.0.0,<7.0.0" },
```

PR #155, opened 2026-06-12
Use structural pattern matching where appropriate

eemeli:
We've required at least Python 3.9 for a while, and so should follow-up on all the
```python
# TODO post-py38: should be a match
```
comments around if...elif...elif...else chains.

### DIFF
--- a/python/moz/l10n/formats/android/parse.py
+++ b/python/moz/l10n/formats/android/parse.py
@@ -498,19 +498,19 @@ def parse_inline(
                     else:
                         # Placeholder
                         func: str | None
-                        # TODO post-py38: should be a match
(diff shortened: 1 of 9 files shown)


mathjazz:
Please fix the failing checks.

eemeli:
Drat, I misremembered that the [match statement](https://docs.python.org/3/reference/compound_stmts.html#match) became available in Python 3.9, and not Python 3.10. So the changes here are blocked by [bug 2014269](https://bugzilla.mozilla.org/show_bug.cgi?id=2014269), "Bump Mach's minimum required Python version to 3.10".

So closing this, and will try again later.

PR #156, opened 2026-06-13
Add `messagesEqual()` JS function

eemeli:
Matches the `==` comparability of Python messages, which we get mostly for free by inheriting from dataclasses.

### DIFF
--- a/js/README.md
+++ b/js/README.md
@@ -25,13 +25,29 @@ and the parser for `mf2` depends on the `messageformat` package.
 
 ## API
 
+### messagesEqual()
+
+```js
+import { messagesEqual } from '@mozilla/l10n'
+```
+
+```ts
+function messagesEqual(a: Message, b: Message): boolean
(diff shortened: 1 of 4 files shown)


mathjazz (commented):
Right now in `messagesEqual()`, if `b` has extra variant keys, those seem to be ignored because the loop only iterates over `av.keys.length`?

Could we compare `av.keys.length` and `bv.keys.length` before comparing individual keys and a regression test for that?

eemeli:
That's not necessary, because https://github.com/mozilla/moz-l10n/blob/36821177461af660d9a81df37117517ab1b29a14/js/src/model-utils.ts#L165-L166 already compares the lengths of the selector arrays, and in a valid message that must match the length of each variant's keys array.

In other words, the comparison presumes that the message are valid.

mathjazz (approved):
Makes sense!

PR #159, opened 2026-06-30
paths: Discover most common locale directory separator

eemeli:
Change required for mozilla/pontoon#4261.

When not using a config file, the locale directory for a new locale with more than one subtag (like `en-GB` or `ca-valencia`) should match the most common pre-existing pattern. The pattern used for the source locale directory is ignored.

When using a config file, this is directly controllable via the `locale_map` argument, for which we provide (and use) `get_android_locale` as a specific implementation.

### DIFF
--- a/python/moz/l10n/paths/discover.py
+++ b/python/moz/l10n/paths/discover.py
@@ -29,11 +29,11 @@ locale_id = re.compile(
 )
 
 
-def dir_contains(dir: str, path: str) -> bool:
+def _dir_contains(dir: str, path: str) -> bool:
     return commonpath((dir, path)) == dir
 
 
-def locale_dirname(base: str, locale: str, locale_dir_sep: Literal["-", "_"]) -> str:
+def _locale_dirname(base: str, locale: str, locale_dir_sep: Literal["-", "_"]) -> str:
     if "-" not in locale or isdir(join(base, locale)):
         return locale
 
@@ -120,7 +120,7 @@ class L10nDiscoverPaths:
         pot_dirs: list[str] = []
         l10n_dirs: list[str] = []
         # Important! We walk with `topdown=True` to get shortest paths first!
-        if ref_root and not dir_contains(root, ref_root):
+        if ref_root and not _dir_contains(root, ref_root):
             walk_roots: Iterator[tuple[str, list[str], list[str]]] = chain(
                 walk(root, topdown=True), walk(ref_root, topdown=True)
             )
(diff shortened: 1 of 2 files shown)


python/moz/l10n/paths/discover.py
```diff
@@ -32,11 +33,14 @@ def dir_contains(dir: str, path: str) -> bool:
     return commonpath((dir, path)) == dir
 
 
-def locale_dirname(base: str, locale: str) -> str:
-    if "-" in locale and not isdir(join(base, locale)):
-        alt_dir = locale.replace("-", "_")
-        if isdir(join(base, alt_dir)):
-            return alt_dir
+def locale_dirname(base: str, locale: str, locale_dir_sep: Literal["-", "_"]) -> str:
```
ewerybody: I thought this might get a default! But I don't see any usage so far. So it's good to be explicit here! 👍

python/moz/l10n/paths/discover.py
```diff
@@ -32,11 +33,14 @@ def dir_contains(dir: str, path: str) -> bool:
...
+    if "-" not in locale or isdir(join(base, locale)):
+        return locale
```
ewerybody, line 38: So `-` is the expected default? Do we have documentation for this?

python/moz/l10n/paths/discover.py
```diff
@@ -191,9 +197,15 @@ def __init__(
         if locale_dirs_:
             self.locales = [dir.replace("_", "-") for dir in locale_dirs_]
             self.locales.sort()
-
+            self.locale_dir_sep = (
+                "_"
+                if sum("_" in dir for dir in locale_dirs_)
+                > sum("-" in dir for dir in locale_dirs_)
+                else "-"
```
ewerybody, line 206: With the amount of occurrences should we have these as CONSTANTS? Maybe like
```
MINUS = "-"
DUNDER = "_"
```

ewerybody (commented):
No complains :)
Just some questions. In general I'd like a line of documentation on each of the the public functions. They're not used outside tho. Should we make them "`_private`"?

python/moz/l10n/paths/discover.py
eemeli: The documentation here is effectively this: https://www.rfc-editor.org/info/rfc5646/#section-2.1

In other words, `-` is the canonically correct separator to use for BCP 47 language tags.

python/moz/l10n/paths/discover.py
eemeli: Eh, I don't think we need that quite yet here, esp. as doing so would add a layer of indirection.

eemeli:
> They're not used outside tho. Should we make them "`_private`"?

Good point; done.

ewerybody (approved):

PR #160, opened 2026-06-30
webext(js): When parsing, match variable names correctly

eemeli:
Fix needed for mozilla/pontoon#4262.

Previously, we were making an incorrect presumption in the JS webext parsing code about how webext variable names are mapped to declaration identifiers. Instead of enforcing them to lower-case, let's actually apply the same mapping we use in Python, and look for a declaration case-insensitively.

### DIFF
--- a/js/src/webext-parse.ts
+++ b/js/src/webext-parse.ts
@@ -14,7 +14,7 @@
  */
 
 import { ParseError } from './errors.ts'
-import type { Message, Pattern } from './model.ts'
+import type { Expression, Message, Pattern } from './model.ts'
 
 export function webextParsePattern(
   msg: Message,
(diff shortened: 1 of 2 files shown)


js/src/webext-parse.ts
```diff
@@ -56,3 +56,11 @@ export function webextParsePattern(
   if (pos < src.length) addText(src.substring(pos))
   return pattern
 }
+
+function findDeclarationName(decl: Record<string, Expression>, name: string) {
```
ewerybody, line 60: In Python this is within the `webext_parse_message` code, right? 
Should we add a reference to that? https://github.com/mozilla/moz-l10n/blob/04e6db742e9b8cb0817f8bae23e85b9c38c80864/python/moz/l10n/formats/webext/parse.py#L104-L106
As it's said in the issue already?

ewerybody (commented):
Seems nice!
If we're doing this to align with Python. Maybe we should have a `find_declaration_name` there as well, doing the same thing?

eemeli:
> If we're doing this to align with Python. Maybe we should have a `find_declaration_name` there as well, doing the same thing?

Probably not, because the Python code is rather different here: The closest corresponding function is `webext_parse_message`, and its second argument is the value of the placeholders object, while in the JS `webextParsePattern` we're getting the data model declarations, which has slightly different considerations.

And so in Python the name mapping is straightforward, while in JS we end up needing to also do a case-insensitive lookup.

ewerybody (commented):
👍

ewerybody (approved):
Ooops yes! That was not an Approve before! Pardon moi!! 🙇

PR #162, opened 2026-07-01
js: Add .properties pattern parser & serialiser

eemeli:
Required for mozilla/pontoon#4270.

Adds a pattern parser & serialiser for .properties files. Does not parse any placeholders or other syntax from the source, so only handles character escapes.


ewerybody:
Ah just saw this! On it! Gonna look into and run it locally as well...

ewerybody (approved):
I don't want to just say "LGTM" ... but I ran it locally. It works 👍 Well, it looks flawless to me! :)

PR #163, opened 2026-07-01
Fix Xliff markup-close serialization

eemeli:
Required for mozilla/pontoon#4266.

On markup-close, the serialiser state was not being correctly set.

The added JS test is just to match Python, and the `repr(Markup)` fix is incidental, as I encountered it while fixing this.


python/moz/l10n/formats/xliff/serialize.py
```diff
@@ -465,7 +465,7 @@ def add_text(text: str) -> None:
                     )
                 if parent.tag != name or parent == el:
                     raise ValueError(f"Improper element nesting for {part} in {parent}")
-                node = parent
+                prev = parent
```
ewerybody, line 468: What a nasty spot to overlook! The new test catches it nicely!👍

ewerybody (approved):
Tested locally making the test catch it. Works nicely 👍

PR #164, opened 2026-07-02
Adding unified cli

### LINKED ISSUE #150: Combine CLI commands under a single `moz-l10n` command
We're starting to accumulate enough CLI commands (thus far: build, build-file, compare, fix, lint, migrate) that they probably ought to be collected under a single `moz-l10n` command, so that they get called as e.g. `moz-l10n build` rather than `l10n-build`.

This should make the commands more discoverable via `moz-l10n --help`.

ewerybody:
Fixes #150
* Moved to `click` for args parsing.
* added some cli helpers to `bin.utils.py`
* moved unit tests into tests/bin
  * now using `pytest` and clicks `CliRunner`
  * new first tests for `fix`, `compare` and `lint`
* added and amended some documentation


python/moz/l10n/bin/build.py
```diff
@@ -173,20 +163,20 @@ def write_target_file(
     total_count = 0
     missing_ids: list[str | list[str]] = []
 
-    def missing_id(id: tuple[str, ...]) -> str | list[str]:
-        # Keep the id structural; a single-part id is simplified to a string.
-        return id[0] if len(id) == 1 else list(id)
+    def missing_id(_id: tuple[str, ...]) -> str | list[str]:
+        """Keep id structural; a single-part id is simplified to a string."""
+        return _id[0] if len(_id) == 1 else list(_id)
```
eemeli: Why `_id` instead of `id` here?

python/moz/l10n/bin/build.py
```diff
@@ -173,20 +163,20 @@ def write_target_file(
...
         nonlocal msg_delta, total_count
         if isinstance(source_entry, Comment):
             return None
         total_count += 1
-        id = section_id + source_entry.id
-        if id in l10n_map:
-            l10n_entry = l10n_map[id]
+        _id = section_id + source_entry.id
+        if _id in l10n_map:
+            l10n_entry = l10n_map[_id]
```
eemeli: Same question as above.

python/moz/l10n/bin/compare.py
eemeli: But this can't be a list?

python/moz/l10n/bin/migrate.py
```diff
@@ -35,72 +33,52 @@
...
-        "--ref",
-        metavar="PATH",
-        type=str,
-        help="path to localization reference root, if separate from --root",
-    )
-    parser.add_argument(
-        "migration", nargs="+", type=str, help="path to migration module(s)"
-    )
-    ns = parser.parse_args(args)
+@click.command(cls=make_list_option_class())
```
eemeli: What does this do? As in, how does this differ from relying on the base `click.Command`?

python/moz/l10n/bin/utils.py
```diff
@@ -0,0 +1,112 @@
...
+
+import click
+
+LOG_FORMAT: str = "%(message)s"
+
+
+def make_list_option_class(
+    option_names: str | list[str] | tuple[str] | None = None,
+    custom_usage: str | None = None,
+) -> type[click.Command]:
```
eemeli: This seems like unnecessary complexity. As we're already changing the CLI to require calling e.g. `moz-l10n compare` instead of `l10n-compare`, we can just as well also start requiring
```
moz-l10n compare --ext ftl,ini
```
instead of
```
l10n-compare --ext ftl ini
```

Then we can also have something simple like
```python
ext_ = set(ext.split(','))
```
within the `cli()` function, rather than needing to introduce `cli_settify` as a separate option.

In general, my preferences will almost always be to do whatever's normal or default, unless there's a really good reason to not do so. Is there a really good reason to not require comma-delimiting the values?

python/tests/bin/test_compare.py
```diff
@@ -0,0 +1,159 @@
...
+
+    with TemporaryDirectory() as tmp_dir:
+        source_json_path = os.path.join(tmp_dir, SOURCE)
+        with open(source_json_path, "w", encoding="utf-8") as f:
+            json.dump(source_map, f)
+
+        fr_dir = os.path.join(tmp_dir, "fr")
+        os.makedirs(fr_dir, exist_ok=True)
+        with open(os.path.join(fr_dir, FILE), "w", encoding="utf-8") as f:
+            f.write(fr_ftl)
```
eemeli: Is there a reason you're not using `build_file_tree` here?

python/tests/bin/test_compare.py
```diff
@@ -0,0 +1,159 @@
...
+        assert "source: 2" in result.output
+
+        result = runner.invoke(moz.l10n.bin.cli, [*cmd_stem, "ini"])
+        assert result.exit_code == 0
+        assert "source: 1" in result.output
+
+        # Test comma separation for multi args via single string
+        cmd = " ".join(cmd_stem) + " .ftl, ini, .txt"
+        # Using single string `shlex.split` is used swallowing win-style double
+        # backslashes m( this works properly on cli tho!
```
eemeli: I don't understand this comment, in particular the `m(`.

Also, in commends and e.g. markdown, I have a strong preference for [semantic line breaks](https://sembr.org/).

python/tests/bin/test_migrate.py
```diff
@@ -0,0 +1,353 @@
...
+from tempfile import TemporaryDirectory
+from textwrap import dedent
+
+import moz.l10n.bin
+import pytest
+from click.testing import CliRunner
+
+from ..utils import Tree, build_file_tree
+
+pytest.importorskip("moz.l10n.formats.android", reason="Requires [xml] extra")
```
eemeli, line 28: Huh, didn't know about this utility. Much better than clobbering it together myself.

python/pyproject.toml
```diff
@@ -31,6 +31,7 @@ classifiers = [
 ]
 requires-python = ">=3.9, <4"
 dependencies = [
+  "click==8.1.7",
```
eemeli: Is our dependency exact, or could this be looser? As in:
```suggestion
  "click ~= 8.1",
```

eemeli (changes requested):
Looks really good! See inline for some comments/questions.

python/moz/l10n/bin/build.py
ewerybody: Oh 🙈 I was playing around with linting options and turned ruff to full blast once. It was highlighting built-ins shadowing for `id` here.

python/moz/l10n/bin/migrate.py
ewerybody: Oh. Probably started off with another one as template and forgot to remove it. If nothing set: This is identical to the original one. Removing it ...

python/moz/l10n/bin/fix.py
```diff
@@ -36,54 +34,38 @@
...
+@click.option("-v", "--verbose", count=True, help="Set logging verbosity")
+@click.option("-q", "--quiet", is_flag=True, help="Only log input argument errors")
+@click.option("--config", metavar="PATH", help="Path to l10n.toml config file")
+@click.option(
+    "--continue",
+    "continue_on_error",
+    is_flag=True,
+    help="Do not stop at first parse error",
+)
+@click.argument("paths", nargs=-1, required=True)
```
flodolo: Isn't `required=True` forcing the user to provide a path? Passing `--config` alone should be a valid use case.

python/moz/l10n/bin/build.py
```diff
@@ -31,58 +32,47 @@
...
-    )
-    logging.basicConfig(format="%(message)s", level=log_level)
-
-    cfg_path: str = args.config
-    l10n_base: str = args.base
-    l10n_target: str = args.target
-    locales: set[str] = set(args.locales)
-    write_coverage: bool = args.coverage
+@click.command(cls=make_list_option_class("--locales"))
+@click.option("--verbose", help="increase logging verbosity")
```
flodolo: Should this have `count=True`, since the `action=count` in the code being replaced?

python/moz/l10n/bin/compare.py
ewerybody: That's right. It cannot!
This is just to satisfy the type checkers. the `compare` func wants `list[str] | set[str]`
https://github.com/mozilla/moz-l10n/blob/7ce6b94480ffdca0bbca41d46e47e93ebdbb5cc5/python/moz/l10n/bin/compare.py#L133
and if you just put `set[str]` it moans like:
```
error[invalid-argument-type]: Argument to function `compare` is incorrect
   --> python/moz/l10n/bin/compare.py:110:35
    |
108 |             continue
109 |         lc = basename(normpath(path))
110 |         errors, missing = compare(source_data, path)
    |                                   ^^^^^^^^^^^ Expected `dict[str, list[str] | set[str]]`, found `dict[str, list[str] | set[str]] | dict[str, set[str]]`
111 |         if json_output:
112 |             json_res[lc] = {
    |
info: element `dict[str, set[str]]` of union `dict[str, list[str] | set[str]] | dict[str, set[str]]` is not assignable to `dict[str, list[str] | set[str]]`
info: └── element `list[str]` of union `list[str] | set[str]` is not assignable to `set[str]`
info: Function defined here
   --> python/moz/l10n/bin/compare.py:132:5
    |
132 | def compare(
    |     ^^^^^^^
133 |     source_data: dict[str, list[str] | set[str]], root: str
    |     -------------------------------------------- Parameter declared here
134 | ) -> tuple[dict[str, str], dict[str, list[str]]]:
135 |     errors: dict[str, str] = {}
    |
```
We could lift the `list[str]` from `compare` but it's also coming in from here as list:
https://github.com/mozilla/moz-l10n/blob/7ce6b94480ffdca0bbca41d46e47e93ebdbb5cc5/python/moz/l10n/bin/compare.py#L78-L79
(well..., json deserialization will never produce a `set` without custom `JSONDecoder`)

python/moz/l10n/bin/compare.py
```diff
@@ -17,75 +15,77 @@
...
-    )
-    parser.add_argument("paths", nargs="+", type=str, help="directories to test")
-    args = parser.parse_args()
+@click.command(cls=make_list_option_class("--ext", custom_usage="PATHS... [OPTIONS]"))
+@click.argument("paths", nargs=-1, required=True)
+@click.option("-v", "--verbose", count=True, help="Set logging verbosity")
+@click.option("--json", "json_output", is_flag=True, help="output JSON")
+@click.option(
+    "--ext",
+    help="File extensions separated by commad or space. Prefix with ! to exclude.",
```
flodolo:
```suggestion
    help="File extensions separated by comma or space. Prefix with ! to exclude.",
```

Are we actually comma-splitting this somewhere?

flodolo (changes requested):
This needs update?
https://github.com/mozilla/moz-l10n/blob/main/python/README.md

python/moz/l10n/bin/utils.py
ewerybody: Hmm I made it extra so that you can do either `--ext ftl,ini` or `--ext ftl ini`.
Originally we have list mangling in each of the commands code. For instance here:
https://github.com/mozilla/moz-l10n/blob/7ce6b94480ffdca0bbca41d46e47e93ebdbb5cc5/python/moz/l10n/bin/build.py#L84
That's where I initially created it for. To make it so that the data already comes into the function as needed and have it modular.

python/tests/bin/test_compare.py
ewerybody: I initially wasn't so familiar with it. I'll look into using it here tho! :)

python/moz/l10n/bin/utils.py
ewerybody: Oh! it works with `--ext ftl, ini` or `--ext ftl ini` not **only** comma separated!

python/tests/bin/test_compare.py
ewerybody: Pardon me. `m(` is an ASCII facepalm. Hmm that was out of frustration over Windows breaking this. We can make it clearer of course! What about:
```
        # Passing `str` instead of `list[str]` into `runner.invoke` makes it use `shlex.split`.
        # It will eat all`os.sep` on Windows breaking the test.
        # This works properly on cli tho!
```
I didn't know about sembr! 😮 that's really useful advise! Before I was just making is so that it not too long lines.

python/tests/bin/test_migrate.py
ewerybody: Also just came across it now :D

python/moz/l10n/bin/fix.py
ewerybody: True! Can do! Pardon me. The test was rather shallow so far.
I will make it run into this explicitly.

python/moz/l10n/bin/compare.py
ewerybody: No. It'd need to be comma AND space. which is crap. I removed it from the help 👍

python/moz/l10n/bin/utils.py
ewerybody: I removed it from help and also testing for comma stuff.

python/moz/l10n/bin/build.py
ewerybody: 0h yes! Fixed all of those, unifying them.

ewerybody: Ohoh. There was actually a bug in there! Fixed and put to the test now! :)

python/moz/l10n/bin/build.py
eemeli: This should still be fixed.

python/moz/l10n/bin/compare.py
eemeli: The better fix here would be to change the `source_data` type to something like `dict[str, Collection[str]]`, and to keep the return type here as `set[str]`.

python/moz/l10n/bin/utils.py
eemeli: I get the point about `cli_settify`, but I would still prefer that we drop this function, and require that the list is either quoted or comma-delimited so that we don't need to extend `click.Command`.

python/moz/l10n/bin/build.py
```diff
@@ -33,7 +33,9 @@
 
 
 @click.command(cls=make_list_option_class("--locales"))
-@click.option("--verbose", help="increase logging verbosity")
+@click.option(
+    "-v", "--verbose", type=int, default=0, help="Set logging verbosity level (0-2)."
+)
```
eemeli, line 49: Previously, the verbosity level was set by repeating the argument multiple times, so that `-v` resulted in a verbosity level 1, while `-vv` resulted in a verbosity level 2. Is something like that supported by click?

eemeli (changes requested):
See replies to previous comments above.

python/tests/bin/test_compare.py
eemeli: Yeah, that's better. Do use "though" rather than "tho", though.

python/moz/l10n/bin/build.py
ewerybody: `entry_id` OK? :)

python/moz/l10n/bin/compare.py
ewerybody: Yeah looking good!

python/moz/l10n/bin/build.py
eemeli: Let's just have `id`; this is very clearly different from the built-in function, and our current ruff config is completely fine with it. If later we want to change our ruff config to be tighter and this does become an issue, then let's have the discussion (again) then.

Also, to avoid shadowing, this [should be `id_`](https://peps.python.org/pep-0008/#:~:text=single%5Ftrailing%5Funderscore%5F) instead of `_id`.

python/moz/l10n/bin/utils.py
ewerybody: Oh these are 2 different things!

* The subclassing is for having:
  * "--options" that have multiple arguments until the next `--option`.
    That was a standard with `argparse` and is here only possible with anonymous `click.argument` which does not have a help string or by multiple instances of `--option` (`multiple=True` i.e. `--option a --option b` -> `option = ('a', 'b')` which I found a bit ridiculous)
  * the possibility to edit the usage-string.
    On `compare` with `--ext` and `paths` it was like it would expect paths to be added to the end, but if `--ext` was the last option it was eating all the paths ...
    Of course hat's solved if we do the extensions differently. I'm fine with `txt,ini,flt` 👍
* the `cli_settify`:
  * to have a set coming right into ... well you know what it's about :)

OK. Then what do we do with `--locales`?
`en,de,nb-NO,it` as well? 👍

python/tests/bin/test_compare.py
ewerybody: This no longer exists 👍

python/moz/l10n/bin/utils.py
eemeli: > OK. Then what do we do with `--locales`?
> `en,de,nb-NO,it` as well? 👍

Yes.

python/moz/l10n/bin/build.py
ewerybody: Ah!!! I see 🤯
How was that done with `--verbose`?

ewerybody: OK sure that's what it does! I ... eh I didn't know that! :D
We need to document this! https://click.palletsprojects.com/en/stable/options/#counting

python/moz/l10n/bin/build.py
ewerybody: OK @eemeli was pointing me to the true meaning of this `count` thing! Fixed and tested for!

python/moz/l10n/bin/compare.py
ewerybody: OK NOW we ARE comma splitting! On these `--ext` args and on `--locales` at the `build` command.

python/moz/l10n/bin/build.py
ewerybody: OK. Fixed and tested for.

python/moz/l10n/bin/build.py
```diff
@@ -31,82 +32,78 @@
...
+
+    For Fluent, adds empty files for any missing from the target locale.
+    For other formats, copies file from the source locale if they are missing from the target.
+    """
+    set_log_level(verbose)
 
     # locale -> [ftl_missing, src_fallback]
     msg_data: dict[str, list[int]] = defaultdict(lambda: [0, 0])
 
+    locales_ = set(locales.split(",") if "," in locales else set([locales]))
```
eemeli: The `if` isn't needed here.
```suggestion
    locales_ = set(locales.split(","))
```

python/moz/l10n/bin/utils.py
```diff
@@ -0,0 +1,82 @@
...
+
+    elif file_paths:
+        root_dir = os.getcwd()
+        path_iter = (path for fp in file_paths for path in glob(fp, recursive=True))
+
+    else:
+        log.error(LOG_ERROR_CONFIG_OR_PATHS)
+        return 2, None, ""
+
+    return 0, path_iter, root_dir
```
eemeli: Two comments:
1. I would find this function easier to read with multiple `return` statements also for success.
2. Given that the error code from here is always `2` on failure and that this corresponds with `path_iter` being empty, the returned tuple should not include the error code. This would also simplify the calling code to only need to check for one thing.

python/tests/bin/test_build.py
```diff
@@ -0,0 +1,447 @@
...
+
+from ..utils import Tree, build_file_tree
+
+CFG_TOML_TEMPLATE = """
+basepath = "."
+locales = ["fr", {}]
+[[paths]]
+    reference = "en/file.ftl"
+    l10n = "{{locale}}/file.ftl"
+"""
```
eemeli: I find adding this kind of indirection to test code to be a net negative, especially as it's not immediately clear that the `{}` is a templating slot, while the `{{locale}}` is not.

Let's just repeat the thing in each test where it's needed, especially as it doesn't really matter that it's the same between tests.

In runtime code, I do not think we ought to copy-paste code around, especially if we end up with more than two repeats.

eemeli (changes requested):
Only some pretty nitpicky comments left! I've also yet to test this manually, which should probably happen before we merge.

python/moz/l10n/bin/build.py
ewerybody: Oh you're right! 🙈 `'x'.split(',') -> ['x']` which is perfectly fine for `set(['x'])`. It was a little late ...

flodolo:
```suggestion
    locales_ = {locale.strip() for locale in locales.split(",")}
```

The ternary should be unnecessary? Also, each element should be stripped.

Technically, I guess you could even do `{locale.strip() for locale in locales.split(",") if locale.strip()}` to avoid empty elements (double comma, trailing comma).

python/moz/l10n/bin/compare.py
```diff
@@ -17,75 +15,82 @@
...
-    if args.ext:
-        arg_ext: list[str] = args.ext
-        if len(arg_ext) == 1 and "," in arg_ext[0]:
-            arg_ext = [ext.strip() for ext in arg_ext[0].split(",")]
-        for ext in arg_ext:
-            if ext.startswith("!"):
-                ext = ext[1:]
-                ext_exclude.add(ext if ext.startswith(".") else f".{ext}")
+    if ext:
+        for item in (ext,) if "," not in ext else ext.split(","):
```
flodolo: Same here: checking for a comma should be unnecessary, and elements in the set should be stripped.

python/moz/l10n/bin/compare.py
```diff
@@ -17,75 +15,82 @@
...
+@click.option(
+    "-v",
+    "--verbose",
+    count=True,
+    help="Increase logging verbosity. (-v/--verbose INFO, -vv DEBUG).",
+)
+@click.option("--json", "json_output", is_flag=True, help="Output JSON.")
+@click.option(
+    "--ext",
+    help="File extension(s). Separate multiple by comma (`ini,flt`). Prefix with ! to exclude (`!json`).",
```
flodolo:
```suggestion
    help="File extension(s). Separate multiple by comma (`ini,ftl`). Prefix with ! to exclude (`!json`).",
```

python/moz/l10n/bin/utils.py
ewerybody: Ah yeah! I now dropped passing the exit_code. But ... I don't know if that's what you're suggesting in 1. 🧐

python/tests/bin/test_build.py
ewerybody: I was looking for the simplest existing pattern and wanted to use this one for the verbosity test introducing a 4th copy of it. But Yeah, I agree, the curly brackets are ffing it up.
I removed the templating, copied it to the only spot where it's modified. So we still have 3 uses. OK?

python/moz/l10n/bin/build.py
ewerybody: Hmm having a ` ` space in there would separate the arguments already but if the user actually puts `""` around it that's `!False`! ☝️🤓 We'd need to strip!
OK, also @eemeli: I re-introduced dedicated callbacks to sort this out right-NEXT to the `cli` functions in `build.py` and `compare.py`. I hope that's alright now? :)

python/moz/l10n/bin/compare.py
ewerybody: see above

python/moz/l10n/bin/compare.py
ewerybody: 🤯 it's actually FTL!!! Like in "Faster Than Light"! Is that an inside joke? :D
Because `fluent` > `fl___t` > `flt` -._/\\,-> `ftl`??!?
However! Fixed! :)

flodolo: https://github.com/projectfluent/fluent/blob/3dbb402ed5af6b64f5c09faeb067acff127a0f34/README.md#fluent-syntax-ftl

python/moz/l10n/bin/build.py
```diff
@@ -31,58 +32,72 @@
...
-
-    cfg_path: str = args.config
-    l10n_base: str = args.base
-    l10n_target: str = args.target
-    locales: set[str] = set(args.locales)
-    write_coverage: bool = args.coverage
+def _locales_settify(
+    context: click.Context, param: click.Parameter, value: str
+) -> set[str]:
+    """Help turning comma separated locales string into proprerly stripped `set`."""
```
flodolo:
```suggestion
    """Help turning comma separated locales string into properly stripped `set`."""
```

python/moz/l10n/bin/lint.py
```diff
@@ -18,80 +16,54 @@
...
+)
+@click.option("-q", "--quiet", is_flag=True, help="Only log input argument errors.")
+@click.option("--config", metavar="PATH", help="Path to l10n.toml config file.")
+@click.option(
+    "-u",
+    "--skip-unknown",
+    is_flag=True,
+    help="Skip files without a known L10n extension.",
+)
+@click.argument("paths", nargs=-1, required=True)
```
flodolo: This is still open. `paths` shouldn't be required.

python/moz/l10n/bin/build.py
```diff
@@ -31,58 +32,72 @@
...
+    is_flag=True,
+    help="Write a coverage.json file per locale with the translation ratio.",
+)
+def cli(
+    verbose: int,
+    config: str,
+    base: str,
+    target: str,
+    locales: set[str],
+    coverage: str | None = None,
```
flodolo: Isn't this a boolean?

python/moz/l10n/bin/build_file.py
```diff
@@ -15,75 +15,64 @@
...
+@click.option("--target", metavar="PATH", required=True, help="Output target.")
+@click.option(
+    "--coverage-base",
+    metavar="DIR",
+    help="Base dir for coverage reporting: update <DIR>/coverage.json with "
+    "this file's translation ratio, keyed by --target's path relative to DIR "
+    "(e.g. browser/browser.ftl), matching l10n-build's keys.",
+)
+def cli(
+    verbose: bool,
```
flodolo: Integer?

python/moz/l10n/bin/lint.py
ewerybody: Oh True! This was still untested. I should probably look into test coverage :|

python/moz/l10n/bin/build.py
ewerybody: True. Fixed.

python/moz/l10n/bin/build_file.py
ewerybody: Darn! I was going over these so often 🙈.

python/moz/l10n/bin/build.py
```diff
@@ -31,58 +32,72 @@
...
+    if not value:
+        return locales
+
+    for item in value.split(","):
+        item = item.strip()
+        if not item:
+            continue
+        locales.add(item)
+
+    return locales
```
eemeli: Wouldn't this also work?
```suggestion
    return set(filter(None, (item.strip() for item in value.split(","))))
```

python/moz/l10n/bin/compare.py
```diff
@@ -17,141 +15,180 @@
...
+            prefix = click.style("  !!!", fg="red", bold=True)
+            click.echo(f"{prefix} {path_}: {error}")
+        if not verbose:
+            continue
+        for path_, messages in missing.items():
+            click.echo(f"  {path_}: {-len(messages)}")
+            if verbose == 1:
+                continue
+            for msg in messages:
+                click.echo(f"    {msg}")
```
eemeli: Any particular reason for preferring `continue` over indentation?

python/moz/l10n/bin/compare.py
```diff
@@ -17,141 +15,180 @@
...
-            except Exception as e:
-                errors[path] = str(e)
+        if not src_messages:
+            continue
+        try:
+            tgt_messages = msg_ids(join(root, path))
+            for msg in src_messages:
+                if msg in tgt_messages:
+                    continue
+                missing.setdefault(path, []).append(msg)
```
eemeli: Here I'd make `missing` a `defaultdict` and:
```suggestion
                if msg not in tgt_messages:
                    missing[path].append(msg)
```

flodolo (approved):
From my point of view (high-level code), this looks good now.

Found another problem (#165), but that's pre-existing.

eemeli (commented):
Oops, looks like I'd forgotten to submit these comments.

python/moz/l10n/bin/compare.py
ewerybody: Oh Yes! Less long arching code. Sorting out things we don't care about early in one place and not at the end again.
And .. well, we're not writing the Linux Kernel here I know but: https://www.kernel.org/doc/html/v4.10/process/coding-style.html#indentation

python/moz/l10n/bin/compare.py
ewerybody: Good idea! Thanks!

python/moz/l10n/bin/build.py
ewerybody: OH I love it! Never worked with `filter` before :O
I added `.strip("\"'")` and now it passes it all! 🚀

python/moz/l10n/bin/compare.py
eemeli: There's a balance here, though. I agree on what you're doing around line 157-158, but for 161-164 the extra `continue` is adding a bit more cognitive load (at least for me) than the extra indent would.

ewerybody: OK. It's also very contained and we're already down from 7 indentations to 5 :)
I changed it back

python/moz/l10n/bin/__init__.py
```diff
@@ -19,6 +19,7 @@
 
 
 @click.group()
+@click.version_option(package_name="moz-l10n")
```
eemeli, line 22: This seemed like a useful addition.

python/moz/l10n/bin/build.py
```diff
@@ -36,7 +36,8 @@ def _locales_settify(
     context: click.Context, param: click.Parameter, value: str
 ) -> set[str]:
     """Help turning comma separated locales string into properly stripped `set`."""
-    return set(filter(None, (item.strip() for item in value.strip("\"'").split(","))))
+    split = value.strip("\"'").split(",")
+    return set(filter(None, (item.strip() for item in split)))
```
eemeli, line 40: There's enough happening here that it's easier to read on two lines.

python/moz/l10n/bin/compare.py
```diff
@@ -181,7 +182,7 @@ def compare(
                 if msg not in tgt_messages:
                     missing[path].append(msg)
         except FileNotFoundError:
-            missing[path].extend(list(src_messages))
+            missing[path] = list(src_messages)
```
eemeli, line 185: If we're in this error handler, `missing[path]` cannot exist, and so extending it would be a bit misleading even if that works.

python/moz/l10n/bin/utils.py
```diff
@@ -52,10 +53,11 @@ def set_log_level(verbose: int, quiet: bool = False) -> int:
 
 def handle_paths(
     config_path: str | None,
-    file_paths: list[str] | tuple[str, ...],
+    file_paths: Sequence[str],
```
eemeli, line 56: Better to extend the type as an ABC.

pyproject.toml
```diff
@@ -2,7 +2,6 @@
 name = "monorepo"
 version = "0.0.0"
 description = "Mozilla tools for localization"
-requires-python = ">=3.9, <4"
```
eemeli, line 5: This is set in `python/pyproject.toml`, we don't need to copy it here.

eemeli (approved):
I played around with the command locally and didn't spot any issues beyond a few help text fixes and a few nitpicks, which seemed better for me to just fix directly.

I also fixed to doc comments to follow [PEP 257](https://peps.python.org/pep-0257/#multi-line-docstrings).

python/moz/l10n/bin/compare.py
```diff
@@ -92,6 +93,7 @@ def cli(
 ) -> None:
     """Compare localizations to their `source` and get a report.
 
+    \b
```
ewerybody, line 96: @eemeli what does this do?

eemeli: It [escapes Click's wrapping](https://click.palletsprojects.com/en/stable/documentation/#escaping-click-s-wrapping).

pyproject.toml
ewerybody: I was adding this when adding `pytest-cov`. [see](https://github.com/mozilla/moz-l10n/pull/164/changes/026bb31c832896d1ff36cca3ab382d96fbe6ac7e)
I don't know what the issue was but that was not a fluke. Lemme check ...

eemeli: The coverage tooling doesn't really have anything to do with the CLI, though. If it's needed, we can do that in a separate PR.

PR #166, opened 2026-07-10
unifying `locale_map` behavior

### LINKED ISSUE #165: build and compare don't resolve {android_locale} in l10n configs
`build` and `compare` crash with `KeyError: 'android_locale'` on any config whose l10n template uses {android_locale}:

```
[[paths]]
    reference = "**/values/strings.xml"
    l10n = "**/values-{android_locale}/strings.xml"
```

Both build a `L10nConfigPaths(source)` without `locale_map` and then call `.format(locale=...)`, so `{android_locale}` is never substituted.

ewerybody:
* removed `migrate` custom handling
* `build`, `compare` and `migrate` now all go through `DEFAULT_LOCALE_MAP` provided by `paths.config`
* added tests for compare and build (migrate test was good already)
* added docs and tests for `android_locale`

fixes #165


python/moz/l10n/paths/config.py
```diff
@@ -31,6 +33,18 @@
 path_stars = compile(r"[*](?:[*](?:[/\\][*]*)?)?")
 path_var = compile(r"{(\w+)}")
 
+DEFAULT_LOCALE_MAP: dict[str, Callable[[str], str]] = {
+    "android_locale": get_android_locale,
+}
```
eemeli: If we make `android_locale` supported by default, then we should probably refactor things so that `locale` is handled through here as well. Or does that make some things messy?

eemeli (changes requested):
Needs a merge from `main`, also see inline.

.github/workflows/docs.yaml
ewerybody, line 1: Oh! darn! That slipped in! I needed it in my main to deploy the docs preview under my account 🙈 There is no conflict though? However I can remove it from the PR...

eemeli: Yes, please leave this out -- it's a separate change.

python/moz/l10n/paths/config.py
```diff
@@ -31,6 +33,22 @@
 path_stars = compile(r"[*](?:[*](?:[/\\][*]*)?)?")
 path_var = compile(r"{(\w+)}")
 
+DEFAULT_LOCALE_MAP: dict[str, Callable[[str], str]] = {
```
eemeli: Is there any need for this to be public? Couldn't we keep it private instead?

Also, isn't the automatic type detection good enough here?

```suggestion
_DEFAULT_LOCALE_MAP = {
```

python/moz/l10n/paths/config.py
```diff
@@ -94,14 +112,17 @@ def __init__(
         To use custom path variables for locales,
         set `locale_map` to be a mapping of path variable names to functions,
         which will be called with `locale` as their only argument.
+        These are merged over `DEFAULT_LOCALE_MAP`, with user entries taking
+        precedence, so the built-in `{locale}` and `{android_locale}` variables
+        are available by default.
```
eemeli:
```suggestion
        These will extend the built-in `{locale}` and `{android_locale}` variables,
        with user entries taking precedence.
```

eemeli (changes requested):
Overall looks good; see inline for nitpicks.

python/moz/l10n/paths/config.py
ewerybody: Yeah, `True` Making it private seems good! We should not advertise amending this to lib users.

But I'd keep the strict typing! It pulls a long trail of `Unknowns` behind it where it would clearly be just `str` with the typing.

eemeli (approved):

PR #171, opened 2026-07-22
Use `@fluent-fn` attribute for original Fluent function names

eemeli:
Required for mozilla/pontoon#4230

This changes the Fluent parsing & serialization to be explicit about which Fluent placeables (i.e. expressions in moz-l10n terms) contain an actual function call (like `{ NUMBER($x) }`), and which do not (`{ $x }`).

During parsing, the Fluent function name is included in the moz-l10n expression as a `@fluent-fn` attribute.

During serialization, unrecognised functions need to have the attribute in order to be serializable. For `:number` and `:datetime` this is optional, though `:number` expressions with no `@fluent-fn` attribute and no options are serialized in Fluent without a `NUMBER()` wrapper.

This also improves the `moz-l10n migrate` behaviour, as it was previously adding `NUMBER()` wrappers in places where we usually do not.

### DIFF
--- a/js/src/fluent-serialize.ts
+++ b/js/src/fluent-serialize.ts
@@ -214,12 +214,9 @@ function expression(
     }
   }
 
-  const arg =
-    '_' in expr && expr._ !== undefined
-      ? literal(expr._)
-      : '$' in expr && isIdentifier(expr.$)
-        ? '$' + expr.$
-        : null
+  let arg: string | undefined
+  if ('_' in expr && expr._ !== undefined) arg = literal(expr._)
+  else if ('$' in expr && isIdentifier(expr.$)) arg = '$' + expr.$
 
   if ('fn' in expr && isIdentifier(expr.fn)) {
     const options: string[] = []
--- a/js/src/fluent-serialize.ts
+++ b/js/src/fluent-serialize.ts
@@ -213,6 +213,14 @@ function expression(
       else decl = declarations[v1]
     }
   }
+
+  const arg =
+    '_' in expr && expr._ !== undefined
+      ? literal(expr._)
+      : '$' in expr && isIdentifier(expr.$)
+        ? '$' + expr.$
+        : null
+
   if ('fn' in expr && isIdentifier(expr.fn)) {
     const options: string[] = []
     if (expr.opt) {
@@ -224,6 +232,7 @@ function expression(
         options.push(`${name}: ${literal(value)}`)
       }
     }
+    let ftlName
     switch (expr.fn) {
       case 'message': {
         if (expr._ !== undefined) {
@@ -249,28 +258,52 @@ function expression(
         const error = 'fluent: Unsupported message or term reference'
         throw new SerializeError(error)
       }
-      case 'number':
-        if (options.length === 0 && isNumber(expr._)) return expr._!
-      // fallthrough
-      default: {
-        if ('_' in expr && expr._ !== undefined) {
-          options.unshift(literal(expr._))
-        } else if ('$' in expr && isIdentifier(expr.$)) {
-          options.unshift('$' + expr.$)
+
+      case 'string':
+        if (!arg) {
+          throw new SerializeError('fluent: Argument required for :string')
         }
-        return expr.fn.toUpperCase() + '(' + options.join(', ') + ')'
+        if (options.length) {
+          const error = 'fluent: Options on :string are not supported'
+          throw new SerializeError(error)
+        }
+        return arg
+
+      case 'datetime':
+      case 'number':
+        if (!arg) {
+          throw new SerializeError(`fluent: Argument required for :${expr.fn}`)
+        }
+        ftlName = expr.attr?.['fluent-fn']
+        if (!ftlName) {
+          if (expr.fn === 'number') {
+            if (options.length === 0) return arg
+            ftlName = 'NUMBER'
+          } else {
+            ftlName = 'DATETIME'
+          }
+        }
+      // fallthrough
+
+      default: {
+        ftlName ??= expr.attr?.['fluent-fn']
+        if (!ftlName || typeof ftlName !== 'string') {
+          const error = `fluent: A @fluent-fn attribute is required for :${expr.fn}`
+          throw new SerializeError(error)
+        }
+        if (arg) options.unshift(arg)
+        return ftlName + '(' + options.join(', ') + ')'
       }
     }
   }
-  if ('_' in expr && expr._ !== undefined) return literal(expr._)
-  if ('$' in expr && isIdentifier(expr.$)) return '$' + expr.$
+  if (arg) return arg
   const error = `fluent: Unsupported pattern part ${JSON.stringify(expr)}`
   throw new SerializeError(error)
 }
 
 function literal(value: string | undefined): string {
   if (!value) return '""'
-  if (isNumber(value)) return value
+  if (/^-?\d+(\.\d*)?$/.test(value)) return value
   const escStr = value
     // eslint-disable-next-line no-control-regex
     .replace(/[\x00-\x1F\\"\x7F-\x9F]/g, (ch) => {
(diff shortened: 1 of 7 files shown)


python/moz/l10n/formats/fluent/serialize.py
```diff
@@ -262,8 +267,9 @@ def fluent_astify_message(
     """
     Transform a message into a corresponding Fluent AST pattern.
 
-    Function names are upper-cased, and expressions using the `message` function
-    are mapped to message and term references.
+    Expressions using the `:message` function are mapped to message and term references.
+    Except for `:datetime` and `:number`, expressions using other functions
+    must have a `@fluent-fn` attribute indicating their corresponding Fluent function name.
```
ewerybody, line 272: Ooof is that identical documentation 6 times in here? Can we make that a little more DRY somehow?🤔

js/src/fluent-serialize.ts
```diff
@@ -213,6 +213,14 @@ function expression(
       else decl = declarations[v1]
     }
   }
+
+  const arg =
+    '_' in expr && expr._ !== undefined
+      ? literal(expr._)
+      : '$' in expr && isIdentifier(expr.$)
+        ? '$' + expr.$
+        : null
```
ewerybody: Bitte WAS?!? Can you add a short comment to this one?

ewerybody (changes requested):
Looking like a pretty neat solution!! 👍
I just have some tiny questions ...

python/moz/l10n/formats/fluent/serialize.py
```diff
@@ -53,8 +53,9 @@ def fluent_serialize(
 
     Section identifiers and multi-part message identifiers are not supported.
 
-    Function names are upper-cased, and expressions using the `message` function
-    are mapped to message and term references.
+    Expressions using the `:message` function are mapped to message and term references.
+    Except for `:datetime` and `:number`, expressions using other functions
+    must have a `@fluent-fn` attribute indicating their corresponding Fluent function name.
```
ewerybody, line 58: that does not apply to `string`?

python/tests/bin/test_migrate.py
ewerybody, line 1: So `migrate` will produce different data now, right? We should put this into the release notes then?

js/src/fluent-serialize.ts
eemeli: Not a fan of ternary, eh? I'll reformulate as a chain of if statements.

python/moz/l10n/formats/fluent/serialize.py
eemeli: It's not mentioned because `:string` does not have a Fluent equivalent, and it's only inserted into the data model to deal with generic selectors. Because of the differences in how the two formats work, there should be no reason for it specifically to ever exist in Fluent.

js/src/fluent-serialize.ts
ewerybody: Not a fan of comments, eh? ;D
Nah I can wrap my head around it just would like to see what's intended easily.

python/moz/l10n/formats/fluent/serialize.py
eemeli: Yes, it's repeated, but because it's a docstring, I don't know that there is any way to avoid the repeat.

python/tests/bin/test_migrate.py
eemeli: Eh, it's mentioned in the PR description, and we don't really have any `moz-l10n migrate` users yet.

js/src/fluent-serialize.ts
eemeli: Correct: Not a fan of comments.

python/moz/l10n/formats/fluent/serialize.py
ewerybody: Well, we don't have any module level docstring yet. (Not many at all actually .. Actually not ANY at all!! 🤯 Is that a design decision?)

ewerybody (approved):

js/src/fluent-serialize.ts
ewerybody: not a `const` anymore right, but who knows what the JS compiler would make out of this anyway? Reads very nicely now!

js/src/fluent.test.ts
```diff
@@ -104,6 +121,7 @@ describe('pattern serialize errors', () => {
     { _: 'foo', fn: 'message', opt: { x: 'y' } }
   ])
   fail('invalid term reference', [{ _: '-foo.bar', fn: 'message' }])
+  fail('function reference with no @fluent-fn', [{ $: 'x', fn: 'foo' }])
```
ewerybody, line 124: Oh! Say do we have this on python side as well now?

flodolo:
Trying to figure out this PR with Claude, it highlighted a possible issue that it's worth confirming.

```
msg =
    { $num ->
        [one] { NUMBER($num) } item
       *[other] { NUMBER($num) } items
    }
```

- Python round-trips this correctly (verified).
- JS throws `SerializeError: Unsupported placeholder {"$":"num","fn":"number","attr":{"fluent-fn":"NUMBER"}} with declaration {"$":"num","fn":"number"}`.

js/src/fluent.test.ts
eemeli: Handling for that, yes: https://github.com/mozilla/moz-l10n/blob/a0921dd1cc14253807231b8f698fd7c299cf0a6e/python/moz/l10n/formats/fluent/serialize.py#L501-L502

But it's not specifically covered by the Python tests.

python/moz/l10n/formats/fluent/serialize.py
```diff
@@ -442,25 +446,39 @@ def expression(
...
     arg: ftl.InlineExpression | None,
-    function: str,
-    options: dict[str, str | VariableRef],
+    expr: Expression,
 ) -> ftl.InlineExpression:
     named: list[ftl.NamedArgument] = []
-    for name, val in options.items():
+    for name, val in expr.options.items():
         ftl_val = value(decl, val)
         if isinstance(ftl_val, ftl.Literal):
```
ewerybody, line 454: We migth change this to
```
if isinstance(ftl_val, (ftl.NumberLiteral, ftl.StringLiteral)):
```
to get rid of some typing mismatch. `fluent.syntax.NamedArgument` expects `value: Union[NumberLiteral, StringLiteral]`
Or rather some other time?

eemeli: I don't see what the typing mismatch here is. `ftl_val` is an `InlineExpression`, which (among other possibilities) can be a `NumberLiteral` or a `StringLiteral`, which are both child classes of `Literal`. So filtering its list of possibilities to `Literal` values leaves us with either a `NumberLiteral` or a `StringLiteral`.

eemeli:
> Trying to figure out this PR with Claude, it highlighted a possible issue that it's worth confirming.

@flodolo Should be fixed now. That message should not really show up ever, but it's of course of the shape that we've been getting because of the parent issue.

python/moz/l10n/formats/fluent/serialize.py
ewerybody: Yes. [`value`](https://github.com/mozilla/moz-l10n/blob/031481127fbdd6a6abf614ba35ec8f22bcfea6ea/python/moz/l10n/formats/fluent/serialize.py#L492-L494) gives us `ftl.InlineExpression` ([via](https://github.com/projectfluent/python-fluent/blob/724a6cf0cba068df12cb43dd9d0071eaec14f18e/fluent.syntax/fluent/syntax/ast.py#L402-L409)) :
```
InlineExpression = Union[
    NumberLiteral,
    StringLiteral,
    MessageReference,
    TermReference,
    VariableReference,
    FunctionReference,
]
```
but `ftl.NamedArgument` accepts only `value: Union[NumberLiteral, StringLiteral]` ([see](https://github.com/projectfluent/python-fluent/blob/724a6cf0cba068df12cb43dd9d0071eaec14f18e/fluent.syntax/fluent/syntax/ast.py#L330)).

ewerybody (approved):
Ok! lets do the typing stuff on the mypy pass.

PR #173, opened 2026-07-24
Add security scanning & SBOM generating GitHub actions

eemeli:
Also configures Dependabot to keep the github-actions updated.

### DIFF
--- a/.github/workflows/security.yaml
+++ b/.github/workflows/security.yaml
@@ -1,0 +1,14 @@
+name: SSDLC Actions
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+
+permissions: {}
+
+jobs:
+  code-security-analysis:
+    permissions:
+      security-events: write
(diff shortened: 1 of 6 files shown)


.github/workflows/security.yaml
```diff
@@ -0,0 +1,14 @@
...
+    branches: [main]
+  pull_request:
+
+permissions: {}
+
+jobs:
+  code-security-analysis:
+    permissions:
+      security-events: write
+    uses: mozilla/ssdlc-actions/.github/workflows/code-security-analysis-public-repo.yml@507c5535545434e6892219c12d8d56a2b39bf710 # v1.1
```
flodolo, line 14: I'm using v1 in VPN and it's working, not sure if it's a bug of v1.1

flodolo: No, v1.1 is working fine there. ~Maybe temporary blip?~

Not a blip, retriggering has the same error. Does it depend on availability of a `dependabot.yml` file?

eemeli: It was the `.*` rule in `.gitignore`.

flodolo (approved):

PR #187, opened 2026-07-27
Moving to `uv_build`

### LINKED ISSUE #185: move to `uv_build`
Since we're leaning into the astral stuff already let's try `uv_build` in a branch and maybe switch to it eventually?

ewerybody:
Fix #185

* removed `setuptools` replaced with `uv_build`
* added `test-wheel.yml` workflow to run our Python unit test suite against the actually installed build.
* wired up the wheel test in `publish.yml` and `python.yml`


eemeli:
A slight bit of scope creep seems appropriate here: Could we also run the test suite against the built version of the package before we push it to pypi?

python/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [build-system]
-requires = ["setuptools >= 61"]
-build-backend = "setuptools.build_meta"
+requires = ["uv_build>=0.11.32,<0.12"]
```
eemeli, line 2: This'll also require an `uv.lock` update.

ewerybody: I can't seem to make it produce any diff there. How would I force it?

python/pyproject.toml
eemeli: The `license` field should probably be updated to:
```suggestion
license = "Apache-2.0"
license-files = ["LICENSE"]
```

python/pyproject.toml
eemeli: Huh, you're right. I guess we're relying on builtin stuff then?

ewerybody: Yes. there is no mention of "setup" or "build" in the lock-file. Seems it's a different pair of socks altogether.

ewerybody:
> A slight bit of scope creep seems appropriate here: Could we also run the test suite against the built version of the package before we push it to pypi?

Hmmm .. I hope you like this one! I'm not a fan of hiding everything in the workflows so this is a new `python/script` file which should work locally **and** in the workflows.

eemeli:
I would actually prefer for the commands to be in a workflow; to me a custom script like this hides things more than workflow commands do.

It seems quite unlikely for issues to ever actually come up with the build, or for them to be platform-specific (esp. given that we do run the tests normally in Linux, MacOS, and Windows), and so we don't really need more than a single job in a well-known environment (i.e. Linux) testing the wheel.

.github/workflows/publish.yaml
eemeli: This should now be restricted to just the `pypi-publish` action.

.github/workflows/test-wheel.yml
```diff
@@ -0,0 +1,59 @@
...
+        run: |
+          mkdir -p ${{ env.TEST_DIR }}
+          cp -r python/tests ${{ env.TEST_DIR }}/
+          cp -r schemas ${{ env.STAGE_DIR }}/
+
+      - name: Prep environment & install wheel
+        working-directory: ${{ env.TEST_DIR }}
+        run: |
+          uv venv .venv --python ${{ matrix.python-version }}
+          source .venv/bin/activate
```
eemeli: This does nothing here.
```suggestion
```

.github/workflows/test-wheel.yml
```diff
@@ -0,0 +1,59 @@
...
+          uv pip install "${{ github.workspace }}/dist/"*.whl pytest jsonschema pytest-subtests
+
+      - name: Run Test
+        working-directory: ${{ env.TEST_DIR }}
+        run: |
+          source .venv/bin/activate
+          pytest -vv
```
eemeli:
```suggestion
        run: uv run pytest -vv
```

.github/workflows/test-wheel.yml
```diff
@@ -0,0 +1,59 @@
...
+          mkdir -p ${{ env.TEST_DIR }}
+          cp -r python/tests ${{ env.TEST_DIR }}/
+          cp -r schemas ${{ env.STAGE_DIR }}/
+
+      - name: Prep environment & install wheel
+        working-directory: ${{ env.TEST_DIR }}
+        run: |
+          uv venv .venv --python ${{ matrix.python-version }}
+          source .venv/bin/activate
+          uv pip install "${{ github.workspace }}/dist/"*.whl pytest jsonschema pytest-subtests
```
eemeli: We don't really need to run the non-xml tests here.
```suggestion
          uv pip install "${{ github.workspace }}/dist/"*.whl[xml]
          uv pip install pytest jsonschema pytest-subtests
```

.github/workflows/test-wheel.yml
```diff
@@ -0,0 +1,59 @@
...
+          source .venv/bin/activate
+          pytest -vv
+
+      - name: Run Test with [xml] extra
+        working-directory: ${{ env.TEST_DIR }}
+        run: |
+          source .venv/bin/activate
+          WHEEL=$(echo "${{ github.workspace }}/dist/"*.whl)
+          uv pip install "${WHEEL}[xml]"
+          pytest -vv
```
eemeli: See suggestion above about always installing with `[xml]`.
```suggestion
```

eemeli (changes requested):

.github/workflows/publish.yaml
ewerybody: Ok. Can I just move and indent it like this?
```
  pypi-publish:
    needs: pypi-test
    runs-on: ubuntu-latest
    environment:
      name: pypi
      url: https://pypi.org/p/moz.l10n
    permissions:
      id-token: write # Required for OIDC
      contents: read
    steps:
      - uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
        with:
          name: python-wheel-artifact
          path: python/dist
      - uses: pypa/gh-action-pypi-publish@ba38be9e461d3875417946c167d0b5f3d385a247 # v1.14.1
        with:
          packages-dir: python/dist

File: .github/workflows/test-wheel.yml (hunk 2 above)
Comment (ewerybody) [reply .github/workflows/test-wheel.yml]: Ah OK. yeah we're not using anything from the venv here?!

File: .github/workflows/test-wheel.yml (hunk 4 above)
Comment (ewerybody) [reply .github/workflows/test-wheel.yml]: Hmm I don't think that works. Using the `*` makes it interpret the `[..]` as regex as well and then it broke like:
```
  source .venv/bin/activate
  uv pip install "/home/runner/work/moz-l10n/moz-l10n/dist/"*.whl[xml]
  pytest -vv
  shell: /usr/bin/bash -e {0}
  env:
    STAGE_DIR: /home/runner/work/_temp/staged_test
    TEST_DIR: /home/runner/work/_temp/staged_test/python
    UV_PYTHON_INSTALL_DIR: /home/runner/work/_temp/uv-python-dir
    UV_PYTHON: 3.9
    UV_CACHE_DIR: /home/runner/work/_temp/setup-uv-cache
error: The wheel filename "*.whl" is invalid: Must have a version
Error: Process completed with exit code 2.
```
[see](https://github.com/mozilla/moz-l10n/actions/runs/30621921526/job/91128282494)
That's why the dance with the `WHEEL` variable but of course I'll remove the non-xml!

File: .github/workflows/test-wheel.yml (hunk 2 above)
Comment (eemeli) [reply .github/workflows/test-wheel.yml]: No, it's not needed because calling `uv` does that automatically.

File: .github/workflows/test-wheel.yml (hunk 4 above)
Comment (eemeli) [reply .github/workflows/test-wheel.yml]: Ah, that makes sense; do that.

File: .github/workflows/publish.yaml
Comment (eemeli) [reply .github/workflows/publish.yaml]: Hang on, we need this for npm as well -- so let's just leave this as-is.

Comment (ewerybody) [reply .github/workflows/publish.yaml]: I added it to the npm job explicitly. OK?

File: .github/workflows/python.yaml [hunk 6]
```diff
@@ -29,6 +29,27 @@ jobs:
...
+      - uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        with:
+          name: python-wheel-artifact
+          path: python/dist/*.whl
+
+  test-package:
+    needs: build-package
+    uses: ./.github/workflows/test-wheel.yml
+    with:
+      artifact-name: python-wheel-artifact
```
eemeli, line 51: See the error here: https://github.com/mozilla/moz-l10n/actions/runs/30626333975

Not sure if the right fix is to add the permission where it's needed (here and in publish.yaml) or drop the permission if it's not needed.

eemeli (changes requested):

.github/workflows/python.yaml
ewerybody: I see! OK I made it `{}` as well 🤞 lets see ...

eemeli (approved):

PR #193, opened 2026-08-11
fluent: Enforce lexicographic order for named options & properties

eemeli:
Ensure that a stable order is used for serialising named arguments and properties in Fluent.

Upcoming Pontoon changes will allow us to switch its resource serialisation to depend on the data model values. While JS & Python are pretty good about keeping these keys in order, [Postgres is not](https://www.postgresql.org/docs/current/datatype-json.html#:~:text=does%20not%20preserve%20the%20order%20of%20object%20keys). So to ensure that the order does not change effectively randomly, let's enforce lexicographic order.

I'll file separate PRs for similar issues in different formats.

### DIFF
--- a/js/src/fluent-serialize.ts
+++ b/js/src/fluent-serialize.ts
@@ -49,7 +49,12 @@ export function fluentSerializeEntry(
   }
   str += '\n'
   if (entry['+']) {
-    for (const [name, value] of Object.entries(entry['+'])) {
+    const properties = Object.entries(entry['+']).sort((a, b) => {
+      const an = a[0].toLowerCase().replace('accesskey', '\uFFFFaccesskey')
+      const bn = b[0].toLowerCase().replace('accesskey', '\uFFFFaccesskey')
+      return an < bn ? -1 : an > bn ? 1 : 0
+    })
+    for (const [name, value] of properties) {
       str += `    .${name} =`
       const attrStr = fluentSerializeMessage(value, options) || '{ "" }'
       str += attrStr.includes('\n')
@@ -229,6 +234,7 @@ function expression(
         options.push(`${name}: ${literal(value)}`)
       }
     }
+    options.sort()
     let ftlName
     switch (expr.fn) {
       case 'message': {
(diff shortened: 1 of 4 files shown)


js/src/fluent-serialize.ts
```diff
@@ -49,7 +49,12 @@ export function fluentSerializeEntry(
   }
   str += '\n'
   if (entry['+']) {
-    for (const [name, value] of Object.entries(entry['+'])) {
+    const properties = Object.entries(entry['+']).sort((a, b) => {
+      const an = a[0].toLowerCase().replace('accesskey', '\uFFFFaccesskey')
```
flodolo, line 54: Should have an explanation for this `\uFFFF` hack somewhere?

flodolo (approved):

python/moz/l10n/formats/fluent/serialize.py
```diff
@@ -209,6 +209,9 @@ def fluent_astify_entry(
         )
         for key, val in entry.properties.items()
     )
+    attributes.sort(
+        key=lambda a: a.id.name.lower().replace("accesskey", "\uffffaccesskey")
```
ewerybody, line 214: This sorts anything containing "accesskey" back. So "my_accesskey" would go behind "my_apples" although it shouldn't. It also catches, though unlikely "accesskeyMuppets" and sorts it to the end.

To match ONLY "accesskey" we could use
```
attributes.sort(
    key=lambda a: (a.id.name.lower() == "accesskey", a.id.name.lower())
)
```

js/src/fluent-serialize.ts
ewerybody: This sorts anything containing "accesskey" back not just exact matches.
Should we address that?

I proposed a solution on the Python side but my JS-Fu is a little too weak to apply the same idea.

python/moz/l10n/formats/fluent/serialize.py
eemeli: This is intentional; the way we use `accesskey` it should be after the content it's an accesskey for. See for example [this message](https://pontoon.mozilla.org/fi/firefox/all-resources/?search_identifiers=true&search=accesskey&string=222955).

js/src/fluent-serialize.ts
eemeli: > Should have an explanation for this `\uFFFF` hack somewhere?

Added.

ewerybody (approved):

PR #194, opened 2026-08-11
webext: Enforce lexicographic order for placeholder definitions

eemeli:
Ensure that a stable order is used for serialising placeholder definitions in web extension resources.

Upcoming Pontoon changes will allow us to switch its resource serialisation to depend on the data model values. While JS & Python are pretty good about keeping these keys in order, [Postgres is not](https://www.postgresql.org/docs/current/datatype-json.html#:~:text=does%20not%20preserve%20the%20order%20of%20object%20keys). So to ensure that the order does not change effectively randomly, let's enforce lexicographic order.


flodolo (approved):

ewerybody (approved):
nice and clean

PR #202, opened 2026-09-01
Add an `editable` option for parsing Android and Xliff patterns in JS

eemeli:
Required for mozilla/pontoon#4339 and mozilla/pontoon#4430.

When we're generating editable content in Pontoon, we use its [editMessageEntry](https://github.com/mozilla/pontoon/blob/main/translate/src/utils/message/editMessageEntry.ts), which does not take into account some of the specifics of how Android and Xliff patterns are then parsed (both using DOMParser's XML support). This is intentional, and should provide a much better translator experience.

Essentially, what we've been doing here thus far is assuming that a pattern is in the representation that'll be used for it in the serialization, which means that for XML formats a bare `&` is escaped as `&amp;` and `<` as `&lt;`, and for Android that significant whitespace is escaped according to one of the ways in which that's supported.

In our Python code this is not an issue, because 1) `lxml` takes care of it for us, and 2) we're not parsing editable contents in Python.

These considerations need to be relaxed when we're parsing input that's coming from a translator. To enable that, this PR is adding an `editable` option that relaxes and changes the parsing as appropriate for these cases. It's possible that there are other considerations we'll find in other formats, but I'm not aware of such at this time.

While working on this I noticed that parsing content like `foo < bar & baz` _isn't_ failing for us in tests like it should, when not using the new option. That's now reported upstream as capricorn86/happy-dom#2338 and commented on in the test files.

CC @flodolo, @camilapedraza


js/src/xliff.test.ts
```diff
@@ -26,12 +26,12 @@ describe('success', () => {
...
     pattern: Pattern,
     exp: string,
-    xliffIsXcode: boolean = false
+    isXcode: boolean = false
   ) =>
     test(name, () => {
       const src = xliffSerializePattern(pattern)
       expect(src).toBe(exp)
-      const res = xliffParsePattern(src, xliffIsXcode)
+      const res = xliffParsePattern(src, { isXcode })
```
ewerybody, line 34: Oh that was me, right? :O

js/src/xml-utils.ts
```diff
@@ -20,6 +20,11 @@ export function appendText(el: Element, str: string) {
   else el.appendChild(new Text(str))
 }
 
+/** Escape `&` and `<` if they don't look like they're XML syntax */
+export function escapeNonSyntaxChars(str: string) {
```
ewerybody, line 24: This is "only" tested implicitly, right? Should we make an explicit test?

eemeli: I generally prefer only testing external APIs, but given the `happy-dom` bug this might indeed be better tested directly.

js/src/xliff.test.ts
eemeli: With just one flag having it directly as an argument was completely fine, but with two we're at a bit more risk that we'll end up with even more.

js/src/xml-utils.ts
eemeli: Done.

ewerybody (approved):

